PR #13, opened 2017-04-16
Support for proxied class

lukas-krecan:


lukas-krecan:
Did you mean something like this? https://github.com/lukas-krecan/ShedLock/pull/13

PR #18, opened 2017-05-08
Added Jedis as provider

clamey:
Here's an attempt at using Redis as a lock provider with Jedis as the client.  Sadly, there is no decent embeddable Redis instance for testing, so in order to run the JedisLockProviderIntegrationTest you'll need Redis installed locally.  Because of this, I've disabled the shedlock-provider-jedis tests for now so it's not required for ShedLock overall.  The general opinion from the Jedis/Redis world is to have a Redis instance run on your CI box, but I'm not sure that's something you want for the ShedLock TravisCI.

FYI, I tried using the spring-data-redis RedisTemplate instead of Jedis, but that client doesn't support the putIfAbsent with an expiration and evaluate a return semantics that I'd like to use.  Plus the Jedis client interface is much more Redis-like and easier to use that the Spring RedisTemplate.


lukas-krecan-lt:
Cool, thanks, I will look at it.

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
```diff
@@ -0,0 +1,119 @@
...
+                String rez = jedis.set(key, buildValue(), "NX", "PX", difference);
+                if (rez != null && "OK".equals(rez)) {
+                    return Optional.of(new RedisLock(key, jedisPool));
+                }
+            }
+        }
+        return Optional.empty();
+    }
+
+    long getDifference(LockConfiguration lockConfiguration) {
```
lukas-krecan-lt: Could you please write some unit tests for this method? I suspect it might get simplified by using Java 8 date features. It's a bit hard to read right now.

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
```diff
@@ -0,0 +1,119 @@
...
+
+        private RedisLock(String key, JedisPool jedisPool) {
+            this.key = key;
+            this.jedisPool = jedisPool;
+        }
+
+        @Override
+        public void unlock() {
+            try (Jedis jedis = jedisPool.getResource()) {
+                jedis.del(key);
```
lukas-krecan: We should not unlock if lockAtLeastFor did not pass. I think we can simplify locking to take only lockAtMostFor into account and move the lockAtLeastFor logic here.

lukas-krecan: Just to explain - lockAtLeastFor is a hack used by people with really short tasks that want to prevent double execution.

shedlock-provider-jedis/src/test/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProviderIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Optional<SimpleLock> lock1 = getLockProvider().lock(configWithShortTimeout);
+        assertThat(lock1).isNotEmpty();
+
+        sleep(5);
+
+        assertUnlocked(configWithShortTimeout.getName());
+    }
+
+    @Override
+    public void shouldLockAtLeastFor() throws InterruptedException {
```
lukas-krecan: Remove this test please, the original one was correct.

lukas-krecan:
Great job, thanks a lot

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
clamey: OK, I'll give it a shot

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
clamey: OK - lemme try that

shedlock-provider-jedis/src/test/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProviderIntegrationTest.java
clamey: Whups!  I was using that for some learning/debugging - I'll take it out - sorry about that.

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
```diff
@@ -0,0 +1,133 @@
...
+        this.atMostHoursDefault = Duration.of(atMostHoursDefault, ChronoUnit.HOURS);
+    }
+
+    @Override
+    public Optional<SimpleLock> lock(LockConfiguration lockConfiguration) {
+        Instant now = Instant.now();
+
+        // Get 'furthest out' configured expire time for lock TTL
+        long expireTime = Math.max(
+            Duration.between(now, lockConfiguration.getLockAtMostUntil()).toMillis(),
```
lukas-krecan: I am just adding check that lockAtMostUntil is always after lockAtLeastUntil so this will not be necessary

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
```diff
@@ -0,0 +1,133 @@
...
+            this.jedisPool = jedisPool;
+            this.lockConfiguration = lockConfiguration;
+        }
+
+        @Override
+        public void unlock() {
+            Instant now = Instant.now();
+            Instant atLeastUntil = lockConfiguration.getLockAtLeastUntil();
+
+            if (now.equals(atLeastUntil) || now.isAfter(atLeastUntil)) {
```
lukas-krecan: Let's imagine I have lockAtleastFor 5 minutes and lockAtMostFor 1 hour and the task takes 2 minutes. We should hold the lock for 3 more minutes.

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
```diff
@@ -0,0 +1,133 @@
...
+            Duration.between(now, lockConfiguration.getLockAtLeastUntil()).toMillis());
+
+        String key = buildKey(lockConfiguration.getName(), this.environment);
+
+        try (Jedis jedis = jedisPool.getResource()) {
+            String rez = jedis.set(key,
+                    buildValue(),
+                    SET_IF_NOT_EXIST,
+                    SET_EXPIRE_TIME_IN_MS,
+                    expireTime < 0 ? atMostHoursDefault.toMillis() : expireTime); // And if not set, use default
```
lukas-krecan: I will also add check that lockAtMostUntil is in the future into the LockConfiguration. If it's not, it's apparently a misconfiguration. We already have default expirationTime defined somewhere else.

clamey: Yea - it was in the Spring code but wasn't making it here.  Makes sense to drop this and use a common one.

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
clamey: Sounds good

shedlock-provider-jedis/src/main/java/net/javacrumbs/shedlock/provider/jedis/JedisLockProvider.java
clamey: Yup - so please feel free to make those changes, as you know both the problem space and rest of the code much better than me.  Thanks for spending time on this, I really appreciate it.

lukas-krecan:
Here are the small changes I made https://github.com/lukas-krecan/ShedLock/commit/6acdd987cfd48e3ae314661e7fd795460e001b87. If it's OK for you, I will release it tomorrow,

PR #29, opened 2017-07-27
Allow any implementation of Pool<Jedis> to be passed to JedisLockProvider

mgrzeszczak:
I was not able to pass JedisSentinelPool as constructor argument to JedisLockProvider, which made it impossible to use it in a Redis slave-master cluster.


lukas-krecan-lt:
Thanks, will release it tonight or tomorrow.

lukas-krecan:
Released as 0.13.0

PR #46, opened 2017-12-22
Readme introduction clarity

zmcmahon-avatria:
I *think* this is supposed to say "one and only one". Or maybe I'm reading it wrong.


lukas-krecan:
Thanks, me english bad :-)

Issue #47, opened 2018-01-18
RedisConnectionFactory support


siposr:
Hello there,

People who use SpringBoot and Redis have some problem with your Redis-provider. SpringBoot gives us some nice and easy configuration with RedisConnectionFactory. This CF uses the same RedisPool inside the implementation but that is private so no way to get it. Instead it gives us RedisConnection.

I created some nice implementation for your LockProvider interface. I can give it back to the community but I need some information about proper contributions here for that.

Thanks in advance.

lukas-krecan:
Hi, thanks, do not hesitate to send a pull-request, contributions are welcome. (current Redis implementation has been contributed as well)

siposr:
Tank you guys, I will see what can I do. :)

siposr:
Main file is uploaded already.
Test file and pom is coming soon.

lukas-krecan:
Cool, let me know if you need any help.

lukas-krecan:
Hi, just in case you do not have time to work on it, I will be more than happy to finish it.

siposr:
I was way too busy sorry. Now I can do some work with this.
It might have one bug but I just can't trace it back. Most of the time it works like charm. Maybe it is my redis setup. Locks go wrong like once in a month.

lukas-krecan:
OK. What sometimes happens is that locked tasks are shorter than the time-shift between machines in the cluster. So the task is executed multiple times (not in parallel but one after each other). Can be prevented by setting lockAtLeastFor. Maybe it's your case, maybe not. In any case, just let me know if you need any help.

siposr:
lockAtLeastFor is set and most of the time it works. Sometimes redis doesn't expire or delete (? under investigate) the key for the lock, so basically it stucks in the db and causes no more job execution.

siposr:
I think we can close it now.

PR #49, opened 2018-01-31
RedisFactoryLockProvider impl

siposr:
#47 for this issue.
For this implementation you need SpringBoot.

Configuration is pretty same as the Jedis version but it uses RedisConnectionFactory instead of JedisPool.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,6 +33,7 @@
         <module>shedlock-provider-jdbc-template</module>
         <module>shedlock-provider-zookeeper-curator</module>
         <module>shedlock-provider-hazelcast</module>
+        <module>shedlock-provider-redis-connectionfactory</module>
     </modules>
 
     <properties>
--- a/shedlock-provider-redis-connectionfactory/pom.xml
+++ b/shedlock-provider-redis-connectionfactory/pom.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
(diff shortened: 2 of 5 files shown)


clamey:
Any chance we could get the main REAME updated to include an example of how to set this up?

siposr:
#### Redis (using RedisConnectionFactory)
Import 
```xml
<dependency>
    <groupId>net.javacrumbs.shedlock</groupId>
    <artifactId>shedlock-provider-redis-connectionfactory</artifactId>
    <version>0.17.1</version>
</dependency>
```

and configure

```java
@Bean
public LockProvider lockProvider(RedisConnectionFactory redisConnectionFactory) {
	return new RedisFactoryLockProvider(redisConnectionFactory);
}
```

siposr:
I dunno what's go wrong with my test, if I run in IDE as JUnit it works. Any idea?

lukas-krecan:
Thanks, I will look at it in next few days.

marcioviegas:
Any chance we get this check failed fixed?

lukas-krecan:
Sorry, I do not understand?

marcioviegas:
Why could I not find that code on master?

lukas-krecan:
It's here https://github.com/lukas-krecan/ShedLock/tree/master/shedlock-provider-redis-spring I have renamed the module. I am still not sure about the name, that's why it's not released yet.

marcioviegas:
Ahhh, ok! I just copied the Provider to my code and I will wait for the release. 
Thanks @lukas-krecan

clamey:
Maybe we change the Jedis one to shedlock-provider-redis-jedis to keep these two consistent like the shedlock-provider-jdbc-* ones?

lukas-krecan:
Yes, but it would require to do something like this https://github.com/lukas-krecan/ShedLock/pull/51 in order to keep it backward-compatible  But maybe better now then later.

lukas-krecan:
Released as 0.18.0

siposr:
You should update the Wiki as well.
Thank you guys for my first contribution. :)

PR #54, opened 2018-02-15
RedisLockProvider made compatible with spring-data-redis 1

lukas-krecan:


providers/redis/shedlock-provider-redis-spring/src/main/java/net/javacrumbs/shedlock/provider/redis/spring/RedisLockProvider.java
```diff
@@ -69,7 +70,11 @@ public RedisLockProvider(RedisConnectionFactory redisConn, String environment) {
     }
 
     private static Expiration getExpiration(Instant until) {
```
clamey: If it's removed from line 60, is getExpiration used anymore?

clamey: nm - I see it now

PR #56, opened 2018-03-02
clearing of LockRegistry

stangls:
Many applications use ShedLock and some database migration tool.

Integration tests for those applications might need cleared and re-migrated databases (and possibly insert fresh data for the test).
When clearing and re-migrating the database however, the shedlock-table is also cleared - which is okay, since the locks should be cleared. But unfortunately, this is not sufficcient, as Shedlock maintains a list of recently created locks in memory. These recently created locks cache is not accessible from the integration tests as they are `private final` fields instantiated without dependency-injection.

Currently, this restriction is bypassed by using reflection during runtime and making the fields accessible.

This PR resolves this problem. It allows to clear the memory-maintained locks without making the `private final` fields accessible.


lukas-krecan:
Cool, thanks

lukas-krecan:
released as 0.18.2

PR #70, opened 2018-06-20
Fixing RedisConnectionFactory sample

alipov:
Using RedisConnectionFactory instead of (perhaps a copy/pasted leftover) JedisPool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,11 +251,12 @@ and configure
 
 ```java
 import net.javacrumbs.shedlock.provider.redis.spring.RedisLockProvider;
+import org.springframework.data.redis.connection.RedisConnectionFactory;
 
 ...
 
 @Bean
-public LockProvider lockProvider(JedisPool jedisPool) {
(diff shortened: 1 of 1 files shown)


alipov:
BTW, the purpose of ENV in this example is not immediately understandable.
I believe we shall either omit it (use a single-param ctor in this example) or explain its purpose. What do you think?

alipov:
Another note that we should add here is that this provider is supported only on spring-data-redis 2.0.0 and above (because it is using `RedisConnection#set` method that returns a Boolean - which was introduced in 2.0.0).
Please let me know if you're OK with it - and if so - I'll add it as well.

lukas-krecan:
Thanks, I am off till Friday. You can add the note, there is one at the end of readme but it wouldn't hurt to add it to Redis section too.

20. 6. 2018 13:48, 13:48, Alex Lipov <notifications@github.com> napsal/a:

PR #77, opened 2018-08-24
be transaction friendly

grmblfrz:
use jdbcTemplate (which internally uses DataSourceUtils.getConnection())
  use @Transactional with Propagation.REQUIRES_NEW

I think using connection.setAutoCommit(true) might lead to problems especially when using JTA transactions, so I propose changing at least JdbcTemplateStorageAccessor to use Springs @Transactional annotation. Additionally, JdbcTemplateLockProvider can use an existing JdbcTemplate.

### DIFF
--- a/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
+++ b/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2009-2017 the original author or authors.
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -19,8 +19,12 @@ import net.javacrumbs.shedlock.core.LockConfiguration;
 import net.javacrumbs.shedlock.support.AbstractStorageAccessor;
 import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.jdbc.core.JdbcTemplate;
-import org.springframework.transaction.annotation.Propagation;
-import org.springframework.transaction.annotation.Transactional;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionDefinition;
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.TransactionCallbackWithoutResult;
+import org.springframework.transaction.support.TransactionTemplate;
 
 import java.sql.Timestamp;
 import java.time.Instant;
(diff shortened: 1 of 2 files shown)


providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
```diff
@@ -15,47 +15,74 @@
...
+	private final String tableName;
+	private final JdbcTemplate jdbcTemplate;
+
+	JdbcTemplateStorageAccessor(JdbcTemplate jdbcTemplate, String tableName) {
+		this.jdbcTemplate = requireNonNull(jdbcTemplate, "jdbcTemplate can not be null");
+		this.tableName = requireNonNull(tableName, "tableName can not be null");
+	}
+
+	@Override
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
```
lukas-krecan: Are you sure that the annotation will get processed? (I am afraid that it will not be scanned as it is in a library package) TransactionalTemplate might be safer choice

providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
```diff
@@ -15,47 +15,74 @@
...
+	public boolean insertRecord(LockConfiguration lockConfiguration) {
+		String sql = "INSERT INTO " + tableName + "(name, lock_until, locked_at, locked_by) VALUES(?, ?, ?, ?)";
+		try
+		{
```
lukas-krecan: Please preserve formatting

providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
grmblfrz: I have changed the code to use TransactionTemplate (added constructor parameter for injecting a PlatformTransactionManager, if not set, a DataSourceTransactionManager is created and used).

providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
grmblfrz: Sorry, should be fixed now.

providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java
```diff
@@ -31,58 +35,72 @@
...
-			statement.setString(3, getHostname());
-			statement.setString(4, lockConfiguration.getName());
-			statement.setTimestamp(5, now);
-		});
-		return updatedRows > 0;
-	}
+    @Override
+    public boolean insertRecord(LockConfiguration lockConfiguration) {
+        String sql = "INSERT INTO " + tableName + "(name, lock_until, locked_at, locked_by) VALUES(?, ?, ?, ?)";
+        return transactionTemplate.execute(status -> {
```
lukas-krecan: I am really sorry, I have mislead you. I am afraid that TransactionTemplate does not create a new transaction. What exactly is the problem with JTA and autocommit?

grmblfrz: TransactionTemplate creates a new transaction if the correct propagationBehavior is st (PROPAGATION_REQUIRES_NEW or PROPAGATION_REQUIRED), so this should work.
The problem with autocommit is that getConnection() possibly returns a connection with a transaction already started - this might commit an already running transaction (just a gut feeling, I could not verify this. But using TransactionTemplate is the clean solution I think.

lukas-krecan:
Hi, it looks great. I have been thinking about a small change like this to keep the original behavior if the user does not specify the transaction manager https://github.com/lukas-krecan/ShedLock/pull/78/commits/b6ea90c934f824ca7d4d413a871b7023357b5be3

lukas-krecan:
Released as 1.1.0

PR #97, opened 2018-10-17
Don't cache MongoLockProvider in the integration test

ddworak:
The properties tested in the abstract test suite should still hold, even when the lock provider is created on demand (it's synonymous to multiple clients trying to acquire the lock). Unfortunately, when only a single instance is used in the test, it's easy to make an uncaught mistake when writing a custom implementation. In my case I've made an error of using regular `set` updates in my `insertRecord` implementation (instead of correct `setOnInsert`) and all tests still passed.


lukas-krecan-lt:
Thanks

ddworak:
@lukas-krecan-lt I have another PR with an actual improvement to the Mongo provider lined up for submission once/if you merge this :)

lukas-krecan:
Hi, I'd like to merge this but tests do not pass in the PR. They are quite fragile, but it seems that your change breaks them (I have no idea why and how). I want to look at it over the weekend.

lukas-krecan:
Released as 2.1.0

PR #101, opened 2018-10-20
Add elasticsearch provider

florian-perrotti:
Compatible with ES 6.4.x version


florian-perrotti:
Hi sure no problem. 
Using the [refresh option](
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html) to force refresh when first indexing and then using the [version conflict](
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update.html) that throws an exception when something has changed so that only index/update is taken into account. 
Thanks

lukas-krecan:
Hi, I have taken liberty to refactor the code and take advantage of Elasticsearch upsert functionality. Please check if it's OK like this https://github.com/lukas-krecan/ShedLock/pull/102. I am still not sure if it's a good idea to use Elasticsearch for locks, but I will leave it to the users if they want to use it or not. (I may add a small disclaimer to the documentation, though)

lukas-krecan:
Moreover, fuzzTestShouldPass sometimes fails on my machine (both with your an mine version) which indicates that there is something wrong. I will look at this.

lukas-krecan:
Can't reproduce fuzzTestShouldPass issues anymore

lukas-krecan:
Released as 2.1.0

PR #104, opened 2018-10-24
Init hostname once

henri-tremblay:
Getting the hostname is quite costly. Especially on slow networks. Since it won't change, caching it seems the good way to do it.


lukas-krecan:
Thanks

PR #108, opened 2018-11-02
Adding Dynamodb lock provider

arbhatt:


providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+
+        @Override
+        public boolean insertRecord(LockConfiguration lockConfiguration) {
+
+            DynamoDB dynamoDB = new DynamoDB(amazonDynamoDB);
+            Table table = dynamoDB.getTable(tableName);
+            UpdateItemSpec spec = new UpdateItemSpec()
+                    .withPrimaryKey(NAME, lockConfiguration.getName())
+                    .withUpdateExpression(
+                            String.format("set %s=:%s, %s=:%s, %s=:%s",
```
lukas-krecan: Can this whole string be a constant?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+                                    .withLong(":" + LOCKED_AT, Instant.now().toEpochMilli()))
+                    .withReturnValues(ReturnValue.ALL_OLD);
+
+            UpdateItemOutcome updateItemOutcome = table.updateItem(spec);
+            return (updateItemOutcome.getItem() == null);
+        }
+
+        private Map<String, AttributeValue> findLockRecord(LockConfiguration lockConfiguration) {
+            DynamoDB dynamoDB = new DynamoDB(amazonDynamoDB);
+            Table table = dynamoDB.getTable(tableName);
```
lukas-krecan: getTable() can be a method (or even a field)

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+    public DynamodbLockProvider(AmazonDynamoDB amazonDynamoDB, String tableName, Duration minLockTime, long concurrentProcessCount) {
+        this(new DynamodbAccessor(amazonDynamoDB, tableName, minLockTime, concurrentProcessCount));
+    }
+
+    DynamodbLockProvider(DynamodbAccessor accessor) {
+        super(accessor);
+    }
+
+    public static class DynamodbAccessor extends AbstractStorageAccessor {
+        private boolean IS_TABLE_EXIST = false;
```
lukas-krecan: Why is this in capital letters?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+            if (!isTableExist(tableName)) {
+                createTable(tableName);
+            }
+        }
+
+        private boolean isTableExist(String tableName) {
+
+            // If table has been created already and we are aware,
+            // then skip the expensive operations
+            if (IS_TABLE_EXIST) {
```
lukas-krecan: This should not be necessary if the createTable method is called only from constructor, we know that it's called only once per instance

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+                return true;
+            }
+
+            // If we are not aware of the table's existance
+            // then find out
+            boolean moreTablesToScan = true;
+            boolean isTableExist = false;
```
lukas-krecan: doesTableExist sounds better

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+
+            DynamoDB dynamoDB = new DynamoDB(amazonDynamoDB);
+            Table table = dynamoDB.getTable(tableName);
+
+            UpdateItemSpec spec = new UpdateItemSpec()
+                    .withPrimaryKey(NAME, lockConfiguration.getName())
+                    .withUpdateExpression(String.format("set %s=:%s", LOCK_UNTIL, LOCK_UNTIL))
+                    .withValueMap(new ValueMap()
+                            .withLong(":" + LOCK_UNTIL,
+                                    Instant.now().isAfter(lockConfiguration.getUnlockTime())
```
lukas-krecan: Why it is not possible to set lockConfiguration.getUnlockTime() in both cases?

README.md
```diff
@@ -168,6 +168,46 @@ public LockProvider lockProvider(MongoClient mongo) {
...
+    AmazonDynamoDBClientBuilder.standard()
+                        .withRegion(dynamodbRegion)
+                        .withCredentials(new com.amazonaws.auth.profile.ProfileCredentialsProvider(profileCredentialProvider))
+                        .build();
+    return new DynamodbLockProvider(amazonDynamoDB);
+}
+
+```
+
+Please note that MongoDB integration requires Mongo >= 2.4 and mongo-java-driver >= 3.4.0
```
lukas-krecan: Copy and paste error

arbhatt: Rightly caught. This line isn't required

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: One benefit on leaving it in the way it is would be a proximity to see the string template and their substitution values. Looking at 
String.format("set %s=:%s, %s=:%s, %s=:%s", LOCKED_BY, LOCKED_BY, LOCK_UNTIL, LOCK_UNTIL, LOCKED_AT, LOCKED_AT)) - the template/placeholder string is "set %s=:%s, %s=:%s, %s=:%s" and they are substituted by LOCKED_BY, LOCKED_BY, LOCK_UNTIL, LOCK_UNTIL, LOCKED_AT, LOCKED_AT. The reader can quickly visualize the update specification being formed here. 

If we make it a constant, then this wouldn't be very obvious as what is happening here. Imagine - String.format(INSERT_LOCK_UPDATE_SPEC, LOCKED_BY, LOCKED_BY, LOCK_UNTIL, LOCK_UNTIL, LOCKED_AT, LOCKED_AT)). This does not make it as clear. The reader needs to refer to the constant definition to grasp what might be going on here. 

A benefit of constant is usually re-use in multiple places so that changes are easy. We are a little localized here within this class. 

Thoughts?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: This is a little remnant from my previous effort when this wasn't part of the constructor. I believe this should not be required any more.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: Agreed

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: This should not be necessary if the createTable method is called only from constructor, so removing this

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: Having table as a field would eliminate the need for creating the object again and again.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+                createTableIfNotExists();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
+        @Override
+        public boolean insertRecord(LockConfiguration lockConfiguration) {
+
+            DynamoDB dynamoDB = new DynamoDB(amazonDynamoDB);
```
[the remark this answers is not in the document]
arbhatt: Since this does not have a conditionalExpression, it behaves like an upsert.

arbhatt:
DynamoDB needs to specify provisioned read and write throughput. I have some a basic calculation on the payload size and with the help of anticipated locking frequency, I am trying to determine the read units and write units to set the provisioned throughputs. You can refer https://github.com/lukas-krecan/ShedLock/pull/108/commits/9e06b2152e61ef5c5de821749e8f208f4e22f5ed#diff-812832120d35a2d9f5bb8fef1d3d9afaR91 for what is going on here.

arbhatt:
Before I push the changes, what is your merge policy? Do you need me to condense my commits into 1 by rewriting history in this branch so that you have one commit underneath your merge? Let me know if I should squash commits or just push them incrementally for these comments?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
lukas-krecan: You are right, setting current time is better, I will fix it in other providers as well

lukas-krecan:
> Before I push the changes, what is your merge policy? Do you need me to condense my commits into 1 by rewriting history in this branch so that you have one commit underneath your merge? Let me know if I should squash commits or just push them incrementally for these comments?

I do not care, just add new commits and I will squash it on merge

providers/dynamodb/shedlock-provider-dynamodb/pom.xml
```diff
@@ -0,0 +1,137 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>shedlock-parent</artifactId>
+        <groupId>net.javacrumbs.shedlock</groupId>
+        <version>2.0.2-SNAPSHOT</version>
```
lukas-krecan: Please, change this version so the build passes (I have made a release since you have forked the project)

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
```diff
@@ -0,0 +1,306 @@
...
+                    moreTablesToScan = false;
+                }
+
+            }
+
+            IS_TABLE_EXIST = isTableExist;
+            return isTableExist;
+        }
+
+        private void createTable(String tableName) throws InterruptedException {
```
lukas-krecan: What about moving this code out of the LockProvider? Either leave the table creation completely on the user (as we do for example in JDBC) or provide a utility method which has to be explicitly called before this LockProvider is used.

On one hand it would make the usage more complicated, on the other hand would decouple the code and force the user to think about quotas and such.

arbhatt: Right now, there are two constructors. 
- public DynamodbLockProvider(AmazonDynamoDB amazonDynamoDB)
- public DynamodbLockProvider(AmazonDynamoDB amazonDynamoDB, String tableName, Duration minLockTime, long concurrentProcessCount)

In former, user needs to specify the handle to dynamo db that they want to use. Table name and the through put us selected as a default. This is the easiest option. 

In the second user needs to specify handle to dynamo db they want to use and in addition to that, name of the table and some parameters around their anticipated use of the locking framework which is then in turn to determine throughput they may require. This gives a little more control to the user in case, they want to use their own table name and know about the locking and unlocking frequency. 

This can be taken to the next level of a power user where they might like to provision tables first and then use this framework to configure with the existing table. This might be suitable for bigger teams where separate group owns and does the infrastructure provisioning work like table creation, bandwidth allocation, etc. The application team would just build the application. Although this can already be achieved using the 2nd constructor if the table already exists; or with the first constructor where the table name happens to be "shedLock", the default name. 

For neatness, a 3rd constructor can be offered - 
- public DynamodbLockProvider(AmazonDynamoDB amazonDynamoDB, String tableName)
This will utilize the already created table; but might create the table with default throughput if does not exist already.
Or, a 4th constructor
public DynamodbLockProvider(AmazonDynamoDB amazonDynamoDB, Table table)
This would provide even greater control to user to even programmatically or manually create a table and pass it on.

To make it even user friendly, we can wrap all these variation with a Builder which can guide the user to the right choice. 

Thoughts?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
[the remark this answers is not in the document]
lukas-krecan: I wanted to say that if it is possible to do upsert, you do not have to implement AbstractStorageAccessor and simplify the code by simply implementing LockProvider

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
lukas-krecan: I would prefer the constant. For me the update expression itself is not of much interest, what I care about are the values of the variables. Moreover it just feels like a waste of processing resources to use String.format every time on something that's basically a constant. of course, it's just a feeling, the cost will be negligible compared to the network call that follows, but anyway.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
lukas-krecan: My concern is more about the design. A class should ideally do only one thing. A LockProvider should just provide a lock and nothing more. Table creation should be concern of some other class. It's not just a theoretical concern, I expect that there will be lot's of different ways how people create their tables and I do not want to prefer one of them before the other. So my favorite solution would be a LockProvider with only one constructor taking Table as an argument and the rest of the code in some other class. May be builder, utility class or what you deem fit. Thanks.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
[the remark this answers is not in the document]
arbhatt: Yes, that looks much simpler implementation.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: If we can have the following somewhere to be referred, should be enough for readability perspective. Creating constants for them by re using existing constants.

 String.format("set %s=:%s, %s=:%s, %s=:%s", LOCKED_BY, LOCKED_BY, LOCK_UNTIL, LOCK_UNTIL, LOCKED_AT, LOCKED_AT))

providers/dynamodb/shedlock-provider-dynamodb/pom.xml
arbhatt: Will do

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamodbLockProvider.java
arbhatt: SRP certainly makes sense and that allows user to create the table in their desired way

lukas-krecan:
@arbhatt Hi, I am truly sorry, I thought you are still working on the PR and was too busy to check. Now we have another PR for DynamodDB #112.  I prefer the other PR a bit more. Since you understand DynamoDB, can you please check #112 and give the author some feedback if needed? Thanks again

PR #112, opened 2018-12-07
Implemented DynamoDBLockProvider

jrehwaldt:
Implemented DynamoDBLockProvider

Characteristics:
* Implementation based off structure and rules from MongoLockProvider
* dates stored as ISO date time String with UTC time (for human readability in the database)
* the required table is *not* auto-created, but assumed to be present
* a programmatic method for table generation is provided as static method
* uses DynamoDbLocal as test environment; requires additional Maven setup


jrehwaldt:
I unfortunately saw only now that there is already pull request #108 
Sorry for this.

There are some notably changes between this implementation and the other one:
* in `lock()`: no delete before updating
* table creation logic is minimal as static method on provider, because many people will use some kind of provisioning anyway; calculating throughput is left to user
* timestamps are stored as _iso string_ instead of _millis in UTC_ for readability in database. Not sure which is preferred.
* tests use assertj-style assertions

Code should be completely in line with `MongoLockProvider`.

I am happy to contribute to the other approach if there are open issues or blockers. Please let me know how you would like to proceed.

jrehwaldt:
I'd consider this PR complete. Let me know what I can improve.

lukas-krecan:
Hi, thanks a lot, I will take a look at it over the weekend

jrehwaldt:
FYI. I added DynamoDB to the TOC and fixed an indentation in DynamoDB's code example in README.md. Nothing else changed during force-push.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
```diff
@@ -0,0 +1,238 @@
...
+     * 2018-12-07T12:30:37.819Z
+     * 2018-12-07T12:30:37.820Z
+     * </pre>
+     *
+     * When using variable fractional digit count as done in {@link DateTimeFormatter#ISO_INSTANT ISO_INSTANT}
+     * and {@link DateTimeFormatter#ISO_OFFSET_DATE_TIME ISO_OFFSET_DATE_TIME} the following sorting
+     * occurs:
+     *
+     * <pre>
+     * 2018-12-07T12:30:37.819Z
```
lukas-krecan: Wow, I was not aware of that.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
```diff
@@ -0,0 +1,238 @@
...
+     * @param dynamodb   DynamoDB to be used
+     * @param tableName  table to be used
+     * @param throughput AWS {@link ProvisionedThroughput throughput requirements} for the given lock setup
+     * @return           a {@link TableDescription table description}
+     *
+     * @throws ResourceInUseException
+     *         The operation conflicts with the resource's availability. You attempted to recreate an
+     *         existing table.
+     */
+    public static TableDescription createLockTable(
```
lukas-krecan: I would prefer to have this in some utility class.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private final String hostname;
+    private final Table table;
+
+    /**
+     * Uses DynamoDB to coordinate locks stored in table {@value DEFAULT_TABLE_NAME}.
+     *
+     * @param dynamodb  DynamoDB to be used
+     */
+    public DynamoDBLockProvider(AmazonDynamoDB dynamodb) {
```
lukas-krecan: I am wondering if the constructor with default table name is useful. In Mongo it is, the collection is created automatically. In JDBC less so, but the excuse is that you create a table by provided DDL statement so it may be useful. But in Dynamo I assume you will create the table using Java code where you can specify some constant so maybe it does not make much sense to have it here.

lukas-krecan: What about adding a constructor with Table as a single parameter?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
```diff
@@ -0,0 +1,238 @@
...
+            String unlockTimeIso = toIsoString(lockConfiguration.getUnlockTime());
+            UpdateItemSpec request = new UpdateItemSpec()
+                    .withPrimaryKey(ID, lockConfiguration.getName())
+                    .withUpdateExpression(RELEASE_LOCK_QUERY)
+                    .withValueMap(new ValueMap()
+                            .withString(":lockUntil", unlockTimeIso)
+                    )
+                    .withReturnValues(ReturnValue.UPDATED_NEW);
+            UpdateItemOutcome updated = table.updateItem(request);
+            assert unlockTimeIso.equals(updated.getItem().getString(LOCK_UNTIL));
```
lukas-krecan: I have never seen assert in real code before. Just out of curiosity, how do you use them?

jrehwaldt: To enable assertion checking just add ‚-ea’ to Java commandline. Intelli and Eclipse usually do this automatically when running tests.

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
jrehwaldt: Fair points. Do you mean DynamoDB‘s Table instance?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
jrehwaldt: What type name would you like to see?

lukas-krecan: It's up to you, I am pretty bad in naming things. LockTableFactory or LockTableCreator?  And what about returning Table instance so we can pass it to the constructor?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
lukas-krecan: Yes, please

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
jrehwaldt: I called it `DynamoDBUtils` for now and left the method static. Also, I return an instance of `Table`. What do you think?

providers/dynamodb/shedlock-provider-dynamodb/src/main/java/net/javacrumbs/shedlock/provider/dynamodb/DynamoDBLockProvider.java
jrehwaldt: I rebased my code on latest master and used `Utils.toIsoString`. Thank you.

jrehwaldt:
Anything else I can do for now to improve the PR? Looking forward to your thoughts. Thank you for this library.

jrehwaldt:
Please rerun the pipeline. I believe it is a hiccup.

lukas-krecan:
Thanks a lot, I will release is in next few days

lukas-krecan:
Released as 2.2.0

PR #114, opened 2018-12-10
DynamoDB: Cleanup imports and add copyright header

jrehwaldt:
Just minor after cleanups of imports and adding Copyright header to Utils class. Forgot those earlier and somehow IDEA didn't pick the imports up.


jrehwaldt:
Follow-up for #112

PR #133, opened 2019-03-12
Finding class level annotation in proxy method mode

volkovs:
Please correct me if I am wrong.

Currently class level SchedulerLock annotation is not found (PROXY_METHOD mode). There are two proxy tests for JDK and class proxies in internal.SpringLockConfigurationExtractorTest. But neither one tests regular Spring bean (implementing interface) to be proxied by ShedLock.


lukas-krecan:
Thanks, looks good

PR #189, opened 2020-01-10
Cassandra Lock Provider

lukas-krecan:


mitjag:
Yes changes are OK and code simplification are great and always welcome. You also updated tests and added CI part - well done. It looks fine to me. Maybe we could also move find method to test class and remove Lock class.

PR #204, opened 2020-02-28
Support custom clock for lock configuration

MagnarAr:
Add support for custom clock for lock configuration as for unit-tests this is the preferred option.

### DIFF
--- a/shedlock-core/src/main/java/net/javacrumbs/shedlock/core/LockConfiguration.java
+++ b/shedlock-core/src/main/java/net/javacrumbs/shedlock/core/LockConfiguration.java
@@ -15,6 +15,7 @@
  */
 package net.javacrumbs.shedlock.core;
 
+import java.time.Clock;
 import org.jetbrains.annotations.NotNull;
 
 import java.time.Instant;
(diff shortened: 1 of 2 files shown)


lukas-krecan:
Thanks, I will look at it over the weekend.

shedlock-core/src/main/java/net/javacrumbs/shedlock/core/LockConfiguration.java
```diff
@@ -36,18 +37,29 @@
...
     private final Instant lockAtLeastUntil;
 
+    private final Clock clock;
+
     public LockConfiguration(@NotNull String name, @NotNull Instant lockAtMostUntil) {
-        this(name, lockAtMostUntil, Instant.now());
+        this(name, lockAtMostUntil, Instant.now(), Clock.systemUTC());
+    }
+
+    public LockConfiguration(@NotNull String name, @NotNull Instant lockAtMostUntil, @NotNull Clock clock) {
```
lukas-krecan: Is this constructor necessary? If not, I would only add the on on line 54.

lukas-krecan:
Can you please describe your use-case? I have been thinking about providing a static method letting users to define the Clock for all possible ShedLock classes. Would it be useful for you?

lukas-krecan:
What about something like this https://github.com/lukas-krecan/ShedLock/pull/205/files#diff-55de1e26a438e5fdbd30324a08ae606f?

shedlock-core/src/main/java/net/javacrumbs/shedlock/core/LockConfiguration.java
MagnarAr: It was added to support similar cases that LockConfiguration already offered, otherwise unnecessary, yes.

MagnarAr:
> What about something like this https://github.com/lukas-krecan/ShedLock/pull/205/files#diff-55de1e26a438e5fdbd30324a08ae606f?

I believe this solves our problem. The basic use-case is that for testing purposes, we would like to mock the clock the same for the whole test-case, the current ShedLock API did not provide solution for that.

But if there is a global ClockProvider which can be configured in the beginning of our unit/component tests, it is just fine.

Will close this PR and waiting for your solution to be published.
Thanks!

lukas-krecan:
Released as 4.5.0

PR #206, opened 2020-03-09
Inject redis template

petebeegle:
Adds overrides to provide a custom `RedisTemplate`. Also exposes a `Builder` to cut down on the number of constructors needed.


lukas-krecan:
Released as 4.5.1, thanks again

PR #222, opened 2020-04-04
Add LockAssert.TestHelper.makeAllAssertsPass

lukas-krecan:


odin568:
Is it intended that the class TestHelper is not public? I cannot directly access it form my UnitTests

lukas-krecan:
Oops, no. Will fix it.

lukas-krecan:
Released as 4.7.1

PR #225, opened 2020-04-10
Implementation of JdbcTemplatePostgresLockProvider

mateuszgrobelny:
Characteristics:
*Lock is based on generic JDBC template lock provider
*Lock uses ON CONFLICT(...) statement to reduce the number of errors in database

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ First of all, we have to import the project
 <dependency>
     <groupId>net.javacrumbs.shedlock</groupId>
     <artifactId>shedlock-spring</artifactId>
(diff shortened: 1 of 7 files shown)


lukas-krecan:
Thanks a lot, it seems to be overkill to duplicate the lock provider, I will try to figure-out how to implement it in current lock provider.

lukas-krecan:
I have introduced Postgres support to standard JdbcTemplateLockProvider (https://github.com/lukas-krecan/ShedLock/pull/228) so this PR is not needed any more. Thanks again.

PR #229, opened 2020-04-11
Migrated original shedlock-provider-dynamodb to shedlock-provider-dyn…

xmrk:
…amodb-legacy.

Re-implemented shedlock-provider-dynamodb to use AWS SDK2 of DynamoDB client.
Changed HsqlJdbcTemplateLockProviderIntegrationTest to use TimeZone.getDefault() as it wasn't working in my timezone.
Upping version to 5.0.0-SNAPSHOT as migrating shedlock-provider-dynamodb and introduction of shedlock-provider-dynamodb-legacy are breaking changes.
Updated README.md with usage for both the newly implemented and legacy DynamoDB providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ executed repeatedly.
   - [JdbcTemplate](#jdbctemplate)
   - [Mongo](#mongo)
   - [DynamoDB](#dynamodb)
+  - [DynamoDB Legacy](#dynamodb-legacy)
   - [ZooKeeper (using Curator)](#zookeeper-using-curator)
   - [Redis (using Spring RedisConnectionFactory)](#redis-using-spring-redisconnectionfactory)
   - [Redis (using Jedis)](#redis-using-jedis)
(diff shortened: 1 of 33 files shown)


lukas-krecan:
Thanks, I will take a look at it. I prefer to maintain backward compatibility so I will most likely keep current dynamo DB module as is and add the new code to a new module. I can take it from here, thanks again.

xmrk:
Yes, I agree on maintaining backwards compatibility but chose to move the original version into dynamodb-legacy and increase the "default" to use AWS SDK2 as I'm assuming that is what people are using from now on.

xmrk:
Also, I didn't realize you were already pulling in testcontainers and I could have used that for the integration test instead of pulling in amazon's LocalDynamoDB code.

lukas-krecan:
It's great as it is, I will try to do it, it will forme me to learn more about the changes.

lukas-krecan:
This PR has been included in PR #230

lukas-krecan:
@xmrk Hi, can you please check PR https://github.com/lukas-krecan/ShedLock/pull/230. I have kept the original module in place as I really do not want to force people to upgrade or to change their code if it's not necessary.

PR #235, opened 2020-04-20
Add Geode Provider

techmachin3:


lukas-krecan:
Thanks a lot, I will look at it in next few days.

providers/geode/shedlock-provider-geode/src/main/java/net/javacrumbs/shedlock/provider/geode/GeodeLockProvider.java
```diff
@@ -0,0 +1,128 @@
...
+        String lockName = lockConfiguration.getName();
+        log.trace("unlock - attempt : {}", lockName);
+        final Instant now = ClockProvider.now();
+        final Instant lockAtLeastInstant = lockConfiguration.getLockAtLeastUntil();
+        if (lockAtLeastInstant.isBefore(now)) {
+            boolean isUnlocked = executeFunction(lockConfiguration,Constants.UNLOCK);
+            log.debug("unlock - done : {}", isUnlocked);
+        } else {
+            try {
+                Thread.sleep(Duration.between(now,lockAtLeastInstant).toMillis());
```
lukas-krecan, line 121: Would it be possible to do this without blocking current thread?

lukas-krecan:
I am not able to run the test. Locally it throws `java.lang.NoClassDefFoundError: org/apache/geode/management/internal/cli/result/ResultData` and on CI it fails like this https://travis-ci.org/github/lukas-krecan/ShedLock/jobs/677620230 Can you please fix it?

providers/geode/shedlock-provider-geode/src/main/java/net/javacrumbs/shedlock/provider/geode/GeodeLockProvider.java
techmachin3: I can , but then unlock would happen in the future and that would have been inconsistent beahviour ( calling app would assume that it is unlocked as it called unlocked ) . Also if i run a async Task then it could throw an error which the calling application won't get . Let me know your thoughts , ideally the leastTime and atMost Behaviour could have been part of the super class which doesnt unlock unless leastTime is passed.

I would ideally want to do away with the leastTime behaviour as systems where the tasks run in millis , the unlock method would unlock only after the task finishes.

lukas-krecan: I admit that the method name is a bit confusing. But other lock providers effectively unlock the task in the future (if `lockConfiguration.getUnlockTime()` is in the future) so this provider should have the same behavior.

techmachin3: Ok done

techmachin3:
I think the build failed as the module wasnt mentioned in pom.xml , wondering how it got picked up for build and test , anyways have updated the blocking unlock to asynchronous unlock.
Have also updated the pom.xml.

Please take a look .

techmachin3:
Ok sorted out the java docs error , checking on the jacoco coverage error ( plugin is saying unable to create coverage , but works fine on local ) , probably as the test starts vm outside of the Testcase , probably its not able to pick up the coverage. 

Also i see we have used jupiter for test rules , somehow intellij doesnt like it and complains for each testSuite that the jupiter-platform jar is unavailable , any suggestions ?

lukas-krecan:
Thank, I will look at it over weekend.

lukas-krecan:
Hi, it seems that the tests still do not work. If I execute `GeodeLockProviderIntegrationTest` it does not run any tests. Can you please fix it? The cause might be that the test is mixing JUnit 4 and 5 annotations.

techmachin3:
Looking at it now .

techmachin3:
Hi Lukas,
The build is failing on travis with the error "maximum" log limit exceed and terminating the jobs.Please can you increase the log limit ? If not possible to increase length , can comment out all other modules

lukas-krecan:
Thanks a lot, will take a look at it

lukas-krecan:
Some of the builds are passing, so feel free to ignore the issue. I will fix it later.

techmachin3:
Ok thanks

Reference tried to fix openJDK issue , https://travis-ci.org/github/lukas-krecan/ShedLock/builds/679451966

However i had to downgrade geode to 1.9.2 , so if we can solve the openJDK issue then please let me know , ideally i would want to use 1.10 as currently 1.12 is latest.

lukas-krecan:
Hi, I am really sorry but Geode is so complex that I am afraid I would not be able to maintain it. I would recommend to create a third party lock provider similar to the one for CosmosDB https://github.com/jesty/shedlock-provider-cosmosdb

techmachin3:
Hi 
Not sure what the issue , jdk8 build is passing and the underlying complexity is same as hazelcast.
If you want I can take a look at openJdK build again.

lukas-krecan:
I managed to get the tests running with test containers but the necessity to deploy the DistributedLockFunction makes automated testing and deployment quite complex. There is nothing like that needed in Hazelcast. I am afraid I will not be able to maintain this code so it would be better if you published it as third party module.

Issue #248, opened 2020-05-15
Support Spring Boot 2.3.x with a reactive MongoClient


codependent:
**Is your feature request related to a problem? Please describe.**
I've upgrade a Spring Boot 2.2.x reactive project to Spring Boot 2.3 and Shedlock cannot be used anymore.

Spring Boot 2.3 only supports Mongo 4.x, and the current `com.mongodb.reactivestreams.client.MongoClient` can't be used to get a database to initialize the `MongoLockProvider` since it doesn't expect a `com.mongodb.reactivestreams.client.MongoDatabase`

**Describe the solution you'd like**
Added support for Mongo 4.x with a reactive MongoClient

lukas-krecan:
Hi, thanks for feedback. Reactive Mongo client is not supported. Feel free to provide a PR. It should be straightforward to create a new LockProvider by copying current Mongo LockProvider and change it to use the reactive driver.

codependent:
@lukas-krecan no problem, I'll prepare it along next week

lukas-krecan:
Released as 4.10.1

PR #249, opened 2020-05-16
Reactive Mongo Support

codependent:
Related issue: https://github.com/lukas-krecan/ShedLock/issues/248

Since `LockProvider` is blocking I implemented it with temporal locking at subscription time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,34 @@ public LockProvider lockProvider(MongoClient mongo) {
 
 Please note that MongoDB integration requires Mongo >= 2.4 and mongo-java-driver >= 3.7.0
 
+
+#### Reactive Mongo
+Import the project
+
+```xml
+<dependency>
+    <groupId>net.javacrumbs.shedlock</groupId>
+    <artifactId>shedlock-provider-reactive-mongo</artifactId>
+    <version>4.9.4</version>
+</dependency>
+```
+
+Configure:
+
+```java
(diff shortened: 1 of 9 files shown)


lukas-krecan:
Thanks, I will take a look at it in next few days

lukas-krecan:
I have taken the liberty to cleanup the code a bit https://github.com/lukas-krecan/ShedLock/pull/250/commits

codependent:
@lukas-krecan Thanks for merging the PR :-)

Just one really small detail: I'd say that wrapping the Publisher in a Supplier (`static <T> T execute(Supplier<Publisher<T>> command) {`) is not necessary and something like this would be  fine:

```
            execute(getCollection().findOneAndUpdate(
                and(eq(ID, lockConfiguration.getName()), lte(LOCK_UNTIL, now)),
                update,
                new FindOneAndUpdateOptions().upsert(true)
            ));
...
```

```
    static <T> T execute(Publisher<T> command) {
        SingleLockableSubscriber<T> subscriber = new SingleLockableSubscriber<>();
        command.subscribe(subscriber);
        subscriber.await();
        Throwable error = subscriber.getError();
        if (error != null) {
            if (error instanceof RuntimeException) {
                throw (RuntimeException) error;
            } else {
                throw new LockException("Error when executing Mongo statement", error);
            }
        } else {
            return subscriber.getValue();
        }
    }
```

lukas-krecan:
you ar right, thanks

PR #319, opened 2020-08-01
Consul lock provider

kaliy:
Hi. I've created a consul lock provider (actually 2 different implementation of consul integration).

I thought about implementing `doExtend` method in SimpleLock implementation but it seems that this functionality is not used in the project besides integration tests.

There are 2 things needs to be mentioned in this PR:

- Consul has minimum 10 seconds session TTL and actually it holds a session for twice of the TTL! The simple implementation of obtaining lock would conflict with the `lockAtLeastFor` low `lockAtMostFor` settings. That's why I've created 2 implementation: one is using standard sessions in consul, the second one handles session renewals and invalidations in background. I've explained the logic and overall idea of both of them in corresponding javadocs.
- The biggest issue with the simple one is that integration test runs for almost a minute :( The reason behind it is that if `lockAtLeastFor` is present and has low value, the minimum session time is 10 seconds. I even had to disable the fuzz test because it takes too much time to run and actually it expects quick unlock.

Please let me know if there is something I can do to improve the PR and the code.


lukas-krecan:
Hi, thanks a lot, it looks good. I do not think we need ConsulSchedulableLockProvider. lockAtMostFor is just a safety net, setting it to less than 20s does not make much sense. To fix lockAtLeastFor we can schedule a task that will unlock the lock and use sessions just for case when the node crashes.

doExtend is used by power users that use just parts of the library, we do not have to support it.

So, if you agree, can you please remove ConsulSchedulableLockProvider and implementing the task for unlocking after lockAtLeastFor? 

Thanks again.

kaliy:
Sure. I will do it this way. I agree that lockAtMostFor is for the exceptional situations.

Anyway, the timeout integration test would take at least 20 seconds but there is no other way to implement it.

Will let you know once I change it.

kaliy:
@lukas-krecan I pushed new changes: now it has just one implementation: ConsulLockProvider

lukas-krecan:
Thanks, I will take a look at it this week. I may do some small cleanup, if you are ok with that. I find it easier than going back and for with code review comments.

kaliy:
sure, no problem

lukas-krecan:
Released as 4.13.0. Thanks

PR #321, opened 2020-08-03
Add shutdown hook for ConsulLockProvider.unlockScheduler

kaliy:
Hi, it seems I missed the shutdown hook in my previous PR. I've added it and also added the close() method that will be automatically detected by spring.

For micronaut it's not the case as it requires an annotation for the destroy method but I don't think it's a good idea to add dependency on `javax.annotation` there so I added a commentary in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -530,13 +530,13 @@ import net.javacrumbs.shedlock.provider.consul.ConsulLockProvider;
 
 ...
 
-@Bean
+@Bean // for micronaut please define preDestroy property @Bean(preDestroy="close")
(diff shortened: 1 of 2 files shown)


lukas-krecan:
Shutdown is one of the weaknesses of ShedLock. The trouble with the shutdown hook is that most likely the ConsulClient will be closed first, so there is no way how to unlock the locks.

kaliy:
ConsulClient doesn't have anything to close and doesn't even have any destroy method so at least with this lock provider it can be safely closed. Not sure about the shutdown hook (will test it today) but with standard spring bean or micronaut disposal mechanism it would work correctly.

lukas-krecan:
Ok, then. Thanks

PR #341, opened 2020-10-21
fix consul session leak

### LINKED ISSUE #340: Consul Provider does not destroy the session when locking was unsuccessful
**Describe the bug**
Using ConsulLocking Provider Version 4.15.0 with Spring.
Configuration:
Using `@EnableSchedulerLock(defaultLockAtMostFor = "PT1h")` on config class and `@SchedulerLock(name = CONSUL_SYNCHRONIZATION_LOCK)` on Scheduled Method

ShedLock logs (lock name redacted):
```
DEBUG n.j.s.p.consul.ConsulLockProvider - Destroying session 87270cbf-844b-c169-c991-b0fa3b53ef1a 
DEBUG n.j.s.c.DefaultLockingTaskExecutor - Task finished, lock '****/Lock' released 
DEBUG n.j.s.p.consul.ConsulLockProvider - Acquired session fefe792b-7766-c07b-5b35-ec41fa701f45 for 3600 seconds 
DEBUG n.j.s.c.DefaultLockingTaskExecutor - Not executing '****/Lock'. It's locked.
```

**Expected behavior**
When the lock could not be acquired, the consul session should be destroyed.

**Actual behavior**
Trying to acquire a Lock opens a new session, when the resources is already locked, the session is never closed by the LockProvider.
Session destruction works fine when the lock could be akquired - the session is destroyed correcty when the lock is being released by the LockProvider.
This causes a huge amount of unnecessary (and - as far as I can see - unobtainable) sessions to be created. These Sessions are also not released on application shutdown either.

haraldpusch:
This fixes #340 

Sadly, the `ConsulLockProviderIntegrationTest` is flaky on my maschine. The `fuzzTestShouldPass` fails most of the time, when I run the complete test class because a single session remains on the consul. When running this particular test alone, everything is fine. I'm not quite sure what the cause of this might be. It feels like race condition, but I didn't find an obvious cause after a quick look over the `FuzzTester` code. Any idea?

The `@AfterEach public void checkSessions()` doesn't really feel great, but I didn't find another way to check this post condition for every test properly.

Additionally, I added some files to the .gitignore (mainly eclipse files) - I hope this is alright. I can revert this change if you don't want this in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,10 @@
 target/
+.classpath
 .idea
 *.iml
+.project
+.settings/
 .gradle/
 build/
+.attach_pid*
 micronaut/micronaut-test/dependency-reduced-pom.xml
--- a/providers/consul/shedlock-provider-consul/src/main/java/net/javacrumbs/shedlock/provider/consul/ConsulLockProvider.java
+++ b/providers/consul/shedlock-provider-consul/src/main/java/net/javacrumbs/shedlock/provider/consul/ConsulLockProvider.java
@@ -100,6 +100,7 @@ public class ConsulLockProvider implements LockProvider, AutoCloseable {
         if (isLockSuccessful) {
             return Optional.of(new ConsulSimpleLock(lockConfiguration, this, sessionId));
         }
+        destroy(sessionId);
         return Optional.empty();
     }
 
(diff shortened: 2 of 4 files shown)


lukas-krecan:
Looks good, thanks again.

lukas-krecan:
Released as 4.15.1

PR #346, opened 2020-11-03
[#345] Log unexpected exceptions at ERROR level

### LINKED ISSUE #345: Warning vs error trying to create a lock
We were caught out last week with tasks failing to execute due to our code constructing a name that was too long.

What surprised is is the this condition didn't bubble up an exception or an ERROR message, rather it behaved in I think the same manner as if no lock was acquired (such as the lock already acquired elsewhere). 

ie we can see this code:

https://github.com/lukas-krecan/ShedLock/blob/master/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java#L71

logs a warning, rather than raising an exception or printing an error.

**Expected behavior**

In our case, we would have preferred an Exception be thrown - we were calling it directly with `executeWithLock()` and an Exception would have made it clearer to us that wasn't working.

Alternatively, if that's too big a change, it would have been helpful to log it at ERROR level. Our logging setup emits metrics by log level, and any emitted ERRORs would have gotten our attention to it much quicker.

**Actual behavior**

Logged at WARN message only. The stack trace is printed, which is helpful, but my feeling is the log level at minimum should be higher.

Happy to submit a PR to change the log message to ERROR. Wanted to check first if this had been considered and rejected before.

(Thanks for a great library!)

ae-govau:
This enables conditions such as a name that is too long to result in an
error instead of a warning being emitted in logs, which can help
operators who alert on the basis of ERRORs. Closes #345.


lukas-krecan:
Thanks

PR #360, opened 2020-12-07
enable advanced consul-provider configuration

DrWifey:
Ready.
I don't know did I fulfill the code-style you follow, but it seems working)

### DIFF
--- a/providers/consul/shedlock-provider-consul/src/main/java/net/javacrumbs/shedlock/provider/consul/ConsulLockProvider.java
+++ b/providers/consul/shedlock-provider-consul/src/main/java/net/javacrumbs/shedlock/provider/consul/ConsulLockProvider.java
@@ -18,6 +18,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
+import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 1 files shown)


DrWifey:
Ready.
I don't know did I fulfill the code-style you follow, but it seems working)

lukas-krecan:
Looks good, thanks. Will release it later this week.

lukas-krecan:
I guess that setting token in setKVValue at al is still missing? And I do not get the DEFAULT_CONSUL_LOCK_POSTFIX, it's backward incompatible change, I would prefer to put there an empty string by default. Is that a problem?

lukas-krecan:
This should fix it https://github.com/lukas-krecan/ShedLock/pull/361

DrWifey:
How is that possible? I didn't modify the default `DEFAULT_CONSUL_LOCK_POSTFIX` value...
https://github.com/DrWifey/ShedLock/blob/shedlock-parent-4.18.0/providers/consul/shedlock-provider-consul/src/main/java/net/javacrumbs/shedlock/provider/consul/ConsulLockProvider.java

lukas-krecan:
My bad, about the default value sorry. I already have PR for the token. Thanks again.

lukas-krecan:
Released as 4.18.0

Issue #366, opened 2020-12-11
Can you support ETCD?


goudai:
Can you support ETCD?

lukas-krecan:
Hi, currently I do not have capacity for that, but you can provide a PR, if you want. It's quite easy, you can get inspiration in one of the other lock providers we have.

grofoli:
@goudai @lukas-krecan I created a pull request for this. @goudai is my implementation something you can make use of?

lukas-krecan:
Released as 4.20.0

PR #371, opened 2021-01-03
https://github.com/lukas-krecan/ShedLock/issues/366 added provider fo…

grofoli:
Implemented a provider for etcd like requested by @goudai in https://github.com/lukas-krecan/ShedLock/issues/366.

Note that etcds [leases only support TTLs in seconds,](https://etcd.io/docs/v3.4.0/dev-guide/interacting_v3/#grant-leases) therefore the integration tests needed some adjustments for this provider.

### DIFF
--- a/providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
+++ b/providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
@@ -35,8 +35,6 @@ import net.javacrumbs.shedlock.core.SimpleLock;
 import net.javacrumbs.shedlock.support.LockException;
 import net.javacrumbs.shedlock.support.annotation.NonNull;
 
-import java.math.BigDecimal;
-import java.math.RoundingMode;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Optional;
@@ -72,7 +70,7 @@ public class EtcdLockProvider implements LockProvider {
     }
 
     private static final class EtcdLock extends AbstractSimpleLock {
-        private static final BigDecimal MILLIS_IN_SECOND = BigDecimal.valueOf(1000);
+        private static final long MILLIS_IN_SECOND = 1000L;
         private final String key;
         private Long successLeaseId;
         private final EtcdTemplate etcdTemplate;
@@ -121,9 +119,7 @@ public class EtcdLockProvider implements LockProvider {
         }
 
         private long getSecondsUntil(Instant instant) {
-            return BigDecimal.valueOf(getMsUntil(instant))
-                .divide(MILLIS_IN_SECOND, RoundingMode.CEILING)
-                .longValue();
+            return (long) Math.ceil((double) getMsUntil(instant) / MILLIS_IN_SECOND);
         }
 
         private long getMsUntil(Instant instant) {
--- a/providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
+++ b/providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
@@ -1,0 +1,198 @@
+/**
+ * Copyright 2009-2021 the original author or authors.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.javacrumbs.shedlock.provider.etcd;
+
+import io.etcd.jetcd.ByteSequence;
+import io.etcd.jetcd.Client;
+import io.etcd.jetcd.KV;
+import io.etcd.jetcd.Lease;
+import io.etcd.jetcd.Txn;
+import io.etcd.jetcd.kv.TxnResponse;
+import io.etcd.jetcd.lease.LeaseGrantResponse;
+import io.etcd.jetcd.op.Cmp;
+import io.etcd.jetcd.op.CmpTarget;
+import io.etcd.jetcd.op.Op;
+import io.etcd.jetcd.options.GetOption;
+import io.etcd.jetcd.options.PutOption;
+import net.javacrumbs.shedlock.core.AbstractSimpleLock;
+import net.javacrumbs.shedlock.core.ClockProvider;
+import net.javacrumbs.shedlock.core.LockConfiguration;
+import net.javacrumbs.shedlock.core.LockProvider;
+import net.javacrumbs.shedlock.core.SimpleLock;
+import net.javacrumbs.shedlock.support.LockException;
+import net.javacrumbs.shedlock.support.annotation.NonNull;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.time.Duration;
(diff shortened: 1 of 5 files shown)


lukas-krecan:
Thanks a lot, I will take a look at it next week.

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
```diff
@@ -0,0 +1,198 @@
...
+
+            // lock at least until is in the past
+            if (keepLockFor <= 0) {
+                try {
+                    etcdTemplate.revoke(successLeaseId);
+                } catch (Exception e) {
+                    throw new LockException("Can not remove node", e);
+                }
+            } else {
+                Long leaseId = etcdTemplate.createLease(keepLockFor);
```
lukas-krecan, line 108: What about the old lease?

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
```diff
@@ -0,0 +1,198 @@
...
+
+                etcdTemplate.updateLease(key, buildValue(), leaseId);
```
lukas-krecan, line 110: Does the method updateLease update lease or the value? Do we really want to update the value?

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
```diff
@@ -0,0 +1,198 @@
...
+                this.successLeaseId = leaseId;
+            }
+        }
+
+        private long getSecondsUntil(Instant instant) {
+            return BigDecimal.valueOf(getMsUntil(instant))
```
lukas-krecan, line 116: `Math.ceil()` would be slightly more effective

lukas-krecan (commented):
Thanks again. As I am not etcd expert, I have few questions

1. Can you please document it. Especially please explain how the leases are used.
2. If I understand the docs correctly, lease TTL is the minumum time, the server can decide to keep the lease longer. Can it really happen?
3. In theory we could implement it without leases, just by using Transactions. The advantage of leases is the keep-alive mechanism, but the code without them might be easier. Did you consider it?

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
grofoli: the lease for the key is updated, similar to the CLI command `etcdctl put foo2 bar2 --lease 694d76c7c330b909`

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
grofoli: OK, I will have to check that. I need solution where I do not lose the fractional digits by dividing a long and casting back to long immediately.

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
grofoli: Should be timed out and then removed, eventually.

grofoli:
@lukas-krecan I am very open to simplify the code. What would be the idea to solve the timeout?

lukas-krecan:
> @lukas-krecan I am very open to simplify the code. What would be the idea to solve the timeout?

I think it's better with leases. The alternative would be to do compare and swap either based on versions or value containing the lock validity. But let's keep the leases.

grofoli:
> I dug deeper to jetcd and if I understand it correctly, for keep-alive to work one has to call `leaseClient.keepAlive`. Is that correct?

I called `keepAlive` in an earlier local version of mine, but after some tests with etcdctl I found out that it should work without calling it.  As far as I remember, the integration test testing the timeout even failed, which might be connected to fact that keepAlive keeps the lease alive forever. 
At least this is how I understood it.

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
grofoli: There is one missing answer, I owe you: whether or not we want to update the lease value. 

Well, this should implement the lockAtLeast functionality. I checked the redis-jedis provider before starting this implementation, and it has a similar implementation AFAIK.

As always, I am open to any suggestions/improvements, though.

grofoli:
> 1. Can you please document it. Especially please explain how the leases are used.
Yes, sure, I will.

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
grofoli: I could change it to `(long) Math.ceil((double) getMsUntil(instant) / 1000);`. Do you like this more than my BigDecimal variant?

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
grofoli: I renamed the method to `putWithLeaseId`, because this is actually what the method does :-)

providers/etcd/shedlock-provider-etcd/src/main/java/net/javacrumbs/shedlock/provider/etcd/EtcdLockProvider.java
lukas-krecan: Yes please

grofoli: ok, for me casting two times looks strange, but I will change it ;-)

grofoli: done

lukas-krecan:
Thanks, I will look at it over the weekend. If you agree, I will do some minor cleanup. I feel it's time consuming to polish the code through PR comments, so I prefer to do minor changes myself. I plan to change the module name to something like `shedlock-provider-etcd-jetcd` as I am quite nervous from the Beta nature of the jetcd API an deeper name structure would allow us to create `etcd-jetcd1` in the future if needed. And maybe I will try to release the first lease, after we create the second one in the unlock method (if it's possible). It bothers me to just leave it there.

grofoli:
> Thanks, I will look at it over the weekend. If you agree, I will do some minor cleanup. I feel it's time consuming to polish the code through PR comments, so I prefer to do minor changes myself. I plan to change the module name to something like `shedlock-provider-etcd-jetcd` as I am quite nervous from the Beta nature of the jetcd API an deeper name structure would allow us to create `etcd-jetcd1` in the future if needed. And maybe I will try to release the first lease, after we create the second one in the unlock method (if it's possible). It bothers me to just leave it there.

Yes, sounds good. If you created the branch **issue-366-etcd-support** in your repository as well, I could switch the PR's target branch from master to issue-366-etcd-support. WDYT? Would this be a good option for you?

lukas-krecan:
I just planned to add commits on top of yours. The way I usually do it is that I create a new PR from your branch. So feel free to do whatever changes you want to this PR and over the weekend I will build on top of it.

lukas-krecan:
Released as 4.20.0

PR #373, opened 2021-01-09
Support for etcd

lukas-krecan:


grofoli:
@lukas-krecan I like your changes, thanks!

Issue #466, opened 2021-03-22
Cassandra error handling


lsj-phse:
Using the Cassandra provider, getting occasional WriteTimeoutException due to LWT.

com.datastax.oss.driver.api.core.servererrors.WriteTimeoutException: Cassandra timeout during CAS write query at consistency SERIAL (4 replica were required but only 3 acknowledged the write)

I'd like to handle the exception where the exceptions come as a warning instead.

Can I handle it from my application instead? How do I do it, in that case? Thank you!

lukas-krecan:
Hi, thanks for feedback. It makes sense, we have something similar in JdbcTemplate lock provider. We can catch Cassandra exception and log it as warning. Do you want to provide a PR or shall I do it?

lsj-phse:
Great! I'll create a PR for this.

lukas-krecan:
Released as 4.22.1

lsj-phse:
Thank you! Would be good to look into your suggestion also, about QueryExecutionException.

PR #471, opened 2021-03-22
#466 Cassandra error handling

lsj-phse:


lukas-krecan:
Thanks a lot, I was thinking that we may catch some exception higher in the hierarchy, like `QueryExecutionException`. What do you think?

lukas-krecan:
Released as 4.22.1

PR #505, opened 2021-04-10
adding micronaut-data lock provider

drmaas:
This is needed when using https://github.com/micronaut-projects/micronaut-data for sql operations, as connections become automatically transaction aware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ executed repeatedly. Moreover, the locks are time-based and ShedLock assumes tha
 + [Components](#components)
 + [Usage](#usage)
 + [Lock Providers](#configure-lockprovider)
+  - [Jdbc](#jdbc)
   - [JdbcTemplate](#jdbctemplate)
+  - [Micronaut Data Jdbc](#micronaut-data-jdbc)
   - [Mongo](#mongo)
   - [DynamoDB](#dynamodb)
   - [DynamoDB 2](#dynamodb-2)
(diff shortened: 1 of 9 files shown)


lukas-krecan:
Hi, thanks a lot, will tak a look at it over the weekend

lukas-krecan:
Hi again, I have few notes

1. The JDBC provider was not documented since I was thinking about decomissioning it. There is JDBC template provider that has more features (like using server time, using upsert on databases that support it, ...) But it requires Spring JDBC template and I understand that it's something you do not want to depend on. I am OK with keeping it, just be aware that there will be no new devlopment in there. Isn't there something like Spring JdbcTemplate in Micronaut that would allow us to use named parameters?
2. If I understand it correctly, the new provider only adds transaction support? Can you please elaborate why is it needed so I understand it?
3. I have a refactoring in mind where we would just wrap the JdbcStorageAccessor in MicronautDataStorageAccessor. It would require extracting something like TransactionTemplate so I will most likely do it myself, I am afraind that I am not be able to explain the idea.

Thanks again.

drmaas:
Thanks @lukas-krecan 

1. Good to know about jdbc. It does work adequately in some of our micronaut applications (that are using jooq instead of micronaut-data). micronaut-data doesn't use spring-like templates - it is more like spring-data or gorm.
2. When using micronaut-data, the connection objects provided by the micronaut datasource, when used directly, and need to be used from within a transaction. Otherwise the locking database queries fail.
3.  #507 looks good to me

lukas-krecan:
Released as 4.23.0

PR #522, opened 2021-04-16
Fix link to liquibase file in README

gedankennebel:


gedankennebel:
The current link in the README for the liquibase file ends up in a 404

lukas-krecan:
Thanks a lot

Issue #606, opened 2021-06-21
Shift lock_until if a task is still running


makcon:
The documentation says:
 
_You have to set lockAtMostFor to a value which is much longer than normal execution time_

but normally it's hard to predict the execution time of a task. I'd suggest to add an optional feature to shit the time in case of a task is still running:

- Before a task will start create a scheduler that runs every n time
- If a task is still running and `lock_until` is about to expire then update the field (`lock_until` + n time)

lukas-krecan:
Hi, thanks for feedback. It's possible but you have to create the lock manually. See this https://github.com/lukas-krecan/ShedLock/issues/443

lukas-krecan:
You can try this (just released) https://github.com/lukas-krecan/ShedLock#extending-the-lock

makcon:
Hi Lukas, as I understood, to use it I have to manually check for a task completion and call the new static method to extend a lock, right? I'm expecting to hide it under library and just use `@SchedulerLock` for scheduled method.
I created my own class `ExtendableLockingTaskExecutor` that extends `DefaultLockingTaskExecutor` and put logic to track a task completion and add some time if it's still running. It's working but I didn't have time to test it carefully. If you are interested I can share my solution.

makcon:
Hi @lukas-krecan, I made the simple change and we are gonna test it. Please take a look https://github.com/lukas-krecan/ShedLock/pull/668/files

PR #668, opened 2021-08-09
#606 periodically extend lock

makcon:


lukas-krecan:
Hi, thanks. I am currently offline, I'll take a look at it in a week.

makcon:
> Hi, thanks. I am currently offline, I'll take a look at it in a week.

Ok, thanks

lukas-krecan:
Hi, I like the idea, but I don't think it should be the default behavior.

1. Not all `LockProviders` support lock extension. Merging this would break code using such providers.
2. In my opinion, most of the users do not need it.

That being said, we can add support for such behavior by creating a `LockProvider` wrapper that would do the same. Users that need the automated extension mechanism would just wrap the original lock providert to the wrapper. What do you think?

makcon:
Hi Lukas, 

1. I'd add a method to the interface to check if a provider supports the extension before creating scheduler
2. Based my experience many developers don't think about that. I saw many situations to solve such problems with the magic number just increasing timeout

I'd make this feature optional and enable it through `LockConfiguration`. If a flag like `extendLock` set to true then create scheduler. And add a new param to spring annotation (or a property). Then users just need to make sure that the provider supports the extension.

lukas-krecan:
Hi, I really do not think it belongs to the ShedLock core. In my opinion it's really a niche use-case and propagating the configuration parameter through the Lock Configuration and annotations is too heavy. I want to keep the core as lean as possible.

Out of curiosity, what is your use-case?

makcon:
There are a lot of use cases. For example, we received a file from a client to process some items. We parsed the file, stored the incoming  data in database. Then we process  a task as a scheduled operation in background. One task can contains various number of items: 1-20K-30K... Only one instance has to take it to avoid unexpected behaviour. So we cannot predict the time of execution, moreover the processing depends on 3rd party calls (calls sometimes can be slow). How to calculate the timeout? Of course, we can set max we have seen, let say 2h. But what if the instance crashes, the scheduled will be locked and nothing happened during the time. 
In one of my previous company we had similar problem with SQS and we solved it creating custom library to periodically extend "in flight" messages in the similar way. It worked like a charm )))

lukas-krecan:
I have been thinking about something like [this](https://github.com/lukas-krecan/ShedLock/pull/672/files#diff-624cef40d11f095f28fa3fa2801939987354f9666d01491269347ce2d10733fcR1) (there are still some edge cases to cover, but it should work)

makcon:
I got your idea, looks good. Then the new `KeepAliveLockProvider` can be defined as spring bean and no need to touch core.

makcon:
Hi @lukas-krecan, during the feature testing in our project I found that using [this way](https://github.com/lukas-krecan/ShedLock/blob/master/shedlock-core/src/main/java/net/javacrumbs/shedlock/support/KeepAliveLockProvider.java#L38) when you accept scheduler outside can be a problem. Clients have to control a pool size and set it >= task count. That means every time when a client add a new scheduled task in a project he has to remember to increase the pool size. As we know it's not possible in reality. I'd suggest to incapsulate it and create the executor with a single thread for each task in `KeepAliveLock`

lukas-krecan:
Hi, unless I am not completely mistaken, I think that `ScheduledExecutorService` can handle much more tasks than it has threads.

makcon:
I did the test:
- created 2 tasks annotated with `@SchedulerLock` and added just `Thread.sleep(55000)` and logging before and after the sleep
- created `Executors.newSingleThreadScheduledExecutor()`
- run 2  instances of the service
- found that one tasks was extended and another one not
- after I set pool size == 2 and it worked as expected

lukas-krecan:
My guess is that you are using the same executor for scheduling the tasks (Spring Boot will automatically pick `ScheduledExecutorService` from Spring contxt as the default scheduler). In that case it of course can handle only as many concurrent tasks as it has threads.

makcon:
Yea, first I created `ScheduledExecutorService` with single thread as a spring bean. And the behavior was exactly you described ))) Then I created the wrapper to set a separate executor and close it properly:  

```java
public static class LockProviderWrapper extends KeepAliveLockProvider implements Closeable {

        private static final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();

        public LockProviderWrapper(MongoTemplate mongoTemplate) {
            super(new MongoLockProvider(mongoTemplate.getDb()), executorService);
        }

        @Override
        public void close() {
            executorService.shutdown();
        }
    }

    @Bean
    public LockProvider lockProvider(MongoTemplate mongoTemplate) {
        return new LockProviderWrapper(mongoTemplate);
    }
```

Now I retested and it works fine. Maybe before I did something wrong ))) 

btw, what do you think if make `KeepAliveLockProvider` implements `Closeable` and close the executor?

Issue #700, opened 2021-09-07
add support for r2dbc 


ricardkollcaku:
**Shedlock currently dont have support for r2dbc drivers.**
Missing support for r2dbc.
There should be created support for r2dbc drivers.

lukas-krecan:
Hi, thanks for feedback. I do not have capacity for that, feel free to send a pull-request.

lukas-krecan:
Released as 4.29.0

szpak:
Nice!

Btw, @lukas-krecan It would be good to mention it in README.

PR #737, opened 2021-10-04
R2dbc provider

sokomishalov:
Draft for #700.
Also, I think it makes sense to provide an alternative reactive core API and implement it for providers which use reactive API (such as Lettuce, Redisson, Mongo, Cassandra). I might try to do it if it's ok for you from an API design perspective.

### DIFF
--- a/providers/r2dbc/shedlock-provider-r2dbc-internal/src/main/java/net/javacrumbs/shedlock/provider/r2dbc/internal/AbstractR2dbcStorageAccessor.java
+++ b/providers/r2dbc/shedlock-provider-r2dbc-internal/src/main/java/net/javacrumbs/shedlock/provider/r2dbc/internal/AbstractR2dbcStorageAccessor.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2009 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package net.javacrumbs.shedlock.provider.r2dbc.internal;
+
+import io.r2dbc.spi.R2dbcDataIntegrityViolationException;
+import io.r2dbc.spi.Statement;
+import net.javacrumbs.shedlock.core.ClockProvider;
+import net.javacrumbs.shedlock.core.LockConfiguration;
+import net.javacrumbs.shedlock.support.AbstractStorageAccessor;
+import net.javacrumbs.shedlock.support.LockException;
+import net.javacrumbs.shedlock.support.annotation.NonNull;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+
+import java.time.Instant;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Internal class, please do not use.
+ */
+public abstract class AbstractR2dbcStorageAccessor extends AbstractStorageAccessor {
+    private final String tableName;
+
+    public AbstractR2dbcStorageAccessor(@NonNull String tableName) {
+        this.tableName = requireNonNull(tableName, "tableName can not be null");
+    }
+
+    @Override
+    public boolean insertRecord(@NonNull LockConfiguration lockConfiguration) {
+        return Boolean.TRUE.equals(Mono.from(insertRecordReactive(lockConfiguration)).block());
+    }
+
+    @Override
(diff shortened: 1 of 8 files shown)


providers/r2dbc/shedlock-provider-r2dbc/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>
+                                net.javacrumbs.shedlock.provider.jdbc
```
lukas-krecan, line 77: Should be changed

lukas-krecan (commented):
Hi, first of all thanks a lot.

1) I do not want to provide an alterantive reactive core API. First of all, I can't imagine how it would help the default Spring or Micronaut implementation. Moreover, once Project Loom is out, people will be removing the ractive layer so I do not want to invest into reactive much.
2) I assume that the only added value is the fact that you already have R2DBC connection pool and you want to use it for ShedLock as well, is that correct? (You don't want to create synchronous JDBC DataSource just for ShedLock)
3) Regarding the implementation. I prefer [JdbcTemplateProvider](https://github.com/lukas-krecan/ShedLock/blob/master/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/JdbcTemplateStorageAccessor.java) since it provides more functionality (using DB time, using upserts etc.) The trouble is, that it is maintainable only thanks to named SQL parameters. I guess that R2DBC does not support named parameters or does it? If not, then support for DB specific code would be too complex, so your approach is ok.

providers/r2dbc/shedlock-provider-r2dbc/pom.xml
sokomishalov: my bad, fixed

sokomishalov:
Thanks for the quick reply and review!
To the points:
1) It would help to use cluster locks in framework-independent reactive-streams pipelines. In spring I suppose we could rework aspects to deal with reactive publishers (just like `@Timed` annotation supports `Mono` and `Flux` now). 
Regarding Loom your point makes sense but is still arguable though. Loom has not been shifted to the latest LTS version (17) and I would not rely on it too much.
2) In my exact case - yes, I would like to reuse the existing r2dbc connection pool.
3) Yes, most r2dbc drivers don't support named parameters yet.

providers/r2dbc/shedlock-provider-r2dbc-internal/src/main/java/net/javacrumbs/shedlock/provider/r2dbc/internal/AbstractR2dbcStorageAccessor.java
```diff
@@ -0,0 +1,136 @@
...
+import java.time.Instant;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Internal class, please do not use.
+ */
+public abstract class AbstractR2dbcStorageAccessor extends AbstractStorageAccessor {
```
lukas-krecan, line 37: What's thereason for extracting it to a new module? I would prefer to keep it package-protected inside the r2dbc module until it's needed elsewhere.

providers/r2dbc/shedlock-provider-r2dbc/src/main/java/net/javacrumbs/shedlock/provider/r2dbc/R2dbcStorageAccessor.java
```diff
@@ -0,0 +1,52 @@
...
+        this.connectionFactory = requireNonNull(connectionFactory, "dataSource can not be null");
+    }
+
+    @Override
+    protected <T> Mono<T> executeCommand(
+        String sql,
+        Function<Statement, Mono<T>> body,
+        BiFunction<String, Throwable, Mono<T>> exceptionHandler
+    ) {
+        return Mono.usingWhen(
```
lukas-krecan, line 44: Wow, that's why I really hope that Project Loom will be a thing. How is anyone supposed to understand code like this? Just ranting, it's not your fault.

providers/r2dbc/shedlock-provider-r2dbc-internal/src/main/java/net/javacrumbs/shedlock/provider/r2dbc/internal/AbstractR2dbcStorageAccessor.java
sokomishalov: I did it for the same reason as jdbc did - to provide the availability to implement lock provider via spring-r2dbc wrapper in another module in future.

providers/r2dbc/shedlock-provider-r2dbc/src/main/java/net/javacrumbs/shedlock/provider/r2dbc/R2dbcStorageAccessor.java
sokomishalov: It's a matter of habit actually. 
And yes, in some cases reactive and FP libraries reinvent a wheel with such constructions (in this case try-with-resources).

PR #740, opened 2021-10-06
R2DBC Lock Provider

lukas-krecan:


lukas-krecan:
@sokomishalov Can you please check why test for other RDBMS are failing? Most of them due to `java.lang.UnsupportedOperationException: Binding parameters is not supported for the statement`

lukas-krecan:
And Oracle blocks indefinitelly

lukas-krecan:
Wow, parameter binding is [vendor specific ](https://r2dbc.io/spec/0.8.6.RELEASE/spec/html/#statements.parameterized). I love such standards

lukas-krecan:
I do not know, can you please check that's indeed the reason. My feeling is that it may be due to some misconfiguration. For example MySql has two Connection implementations and for some reason it's using the wrong one. Unfortunately, I did not have time to dive deeper.

PR #768, opened 2021-10-27
add kubernetes lock provider

boxcee:
So I was thinking to implement a lock provider using Kubernetes ConfigMaps. What do you think? Still need to get the fuzzy test running though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,7 @@
         <module>providers/mongo/shedlock-provider-mongo-reactivestreams</module>
         <module>providers/neo4j/shedlock-provider-neo4j</module>
         <module>providers/elasticsearch/shedlock-provider-elasticsearch</module>
+        <module>providers/kubernetes/shedlock-provider-kubernetes</module>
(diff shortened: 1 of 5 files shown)


lukas-krecan:
Hi, to be honest, I would be afraid to merge the code. First of all, I do not think that using Kubernetes is a good fit for ShedLock. The purpose of K8S is IMHO different. Morevoer, the code contains lot of different timeouts and for me it's hard to reason about. 

If I am not mistaken, K8S is using Zookeeper and we actually have a Zookeeper provider. Wouldn't be better to use that?

I would prefer if you can expose the provider as a third-party module, the same way [Cosmos DB provider is done](https://github.com/lukas-krecan/ShedLock#cosmosdb).

boxcee:
Hey @lukas-krecan, sure, making a third-party module is also okay to be honest :-D

Here is the official go implementation of leader election in k8s: https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go. The process for locking would be similar, wouldn't it?

But anyway, if you think this shouldn't go into your code base I will put it in a third-party module. Will let you know then.

PR #810, opened 2021-12-06
In memory lock privider

lukas-krecan:


kkocel:
@lukas-krecan When you plan to release new version with this change applied?

lukas-krecan:
Just released 4.30.0

PR #990, opened 2022-04-13
README updated

MappyLiam:
@EnableSchedulerLock(defaultLockAtMostFor = "10m") makes an error as below
```
Caused by: java.time.format.DateTimeParseException: Text cannot be parsed to a Duration
```
So changed it to conform to the format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Now we need to integrate the library with Spring. In order to enable schedule lo
 ```java
 @Configuration
 @EnableScheduling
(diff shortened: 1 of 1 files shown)


lukas-krecan:
Hi, can you please describe where it does not work (version, framework). It should [work](https://github.com/lukas-krecan/ShedLock/commit/15302b8fe5530369c7a001c1950b0498d1d99325).

MappyLiam:
Oh. My bad. I used 2.x version. 
Thanks for your fast answer! I'll close the comment :)

PR #1023, opened 2022-05-12
Allow HazelcastLockProvider to reconnect to the client if the connect…

spartanhooah:
Let clients provide a `Supplier`, which `HazelcastLockProvider` can use to refresh the connection to Hazelcast in the event that the connection dies.

### DIFF
--- a/providers/hazelcast/shedlock-provider-hazelcast4/pom.xml
+++ b/providers/hazelcast/shedlock-provider-hazelcast4/pom.xml
@@ -30,6 +30,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>4.4.0</version>
+            <scope>test</scope>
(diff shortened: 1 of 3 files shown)


lukas-krecan:
Hi, thanks a lot for the PR. I am not a Hazelcast expert, but I would expect that the reconnections should be done by the driver. Can you please explain why setting[ reconnection mode](https://stackoverflow.com/a/48059594/277042) is not sufficent?

spartanhooah:
I wasn't aware of that configuration option. I'll see if I can figure out
how to incorporate it into my company's code.

spartanhooah:
The client configuration is provided by our infrastructure team, but I've asked if the reconnection mode is something they'd be willing to add to the provided XML. That being said, I don't see a way to use the provided client with your alternate approach, although it's probably more useful in the wider world.

lukas-krecan:
Hi, as I think the issue in your configuration I will not merge your PR. If you want to do the reconnect manually, you can wrap `HazelcastInstance` (in similar way as is done in HazelcastClientProxy). In getMap() method you can implement your reconnect.

PR #1037, opened 2022-05-23
Use shedlock-parent as BOM

KENNYSOFT:
Mainly references springdoc/springdoc-openapi#1405

If I use this library then I should declare two dependencies like below:

```groovy
implementation 'net.javacrumbs.shedlock:shedlock-spring:4.35.0'
implementation 'net.javacrumbs.shedlock:shedlock-provider-jdbc-template:4.35.0'
```

But there are shedlock-parent artifact that can work as BOM so we can use like:

```groovy
implementation platform('net.javacrumbs.shedlock:shedlock-parent:4.36.0')
implementation 'net.javacrumbs.shedlock:shedlock-spring'
implementation 'net.javacrumbs.shedlock:shedlock-provider-jdbc-template'
```

(Since I'm not using maven, I have no idea how maven users can utilize this functionality)

Also, I think showing child artifacts in https://mvnrepository.com or somewhere is greater than only in documents or searching single artifact from the repository.

* Ref: https://mvnrepository.com/artifact/org.springdoc/springdoc-openapi/1.6.3 (search for 'springdoc-openapi-' to see child artifacts)
* Current: https://mvnrepository.com/artifact/net.javacrumbs.shedlock/shedlock-parent/4.35.0


KENNYSOFT:
I just found that hazelcast <= 3 support is dropped, should I also remove it from dependencyManagement? Maybe from README also?

lukas-krecan:
Thanks a lot, I will look at the PR in next few weeks. 

If you can send a separate PR with hazelcast 3 removal, I will be grateful.

KENNYSOFT:
hazelcast 3 removal have already happened in 4.34.0, so I think you meant to separate the README updates. Is it right?

lukas-krecan:
Yes, thanks

lukas-krecan:
Regarding the BOM, almost all users will only import 2 ShedLock dependencies. In such case I find using a variable for version easier. Why is a BOM better for you?

KENNYSOFT:
Right, I can use a variable, but this makes dependabot cannot recognize the version since it just tries to match regex for dependency lines. (and it's so hard that make dependabot to analyze executable scripts)

If dependabot makes two PRs for adjacent lines, merging one of it always result to conflict. :cry:

lukas-krecan:
You are right about the dependabot, I have never noticed that. I will prefer to have a separate module for the BOM, it seems like a misuse of the parent project. I will look at it.

lukas-krecan:
Would soemthing like [this](https://github.com/lukas-krecan/ShedLock/pull/1045) work for you?

KENNYSOFT:
Yeah, cool!

I've built (maven lifecycle `install`) it in local with version 4.34.0 for a test.

By declaring `mavenLocal()` dependencies, it perfectly resolved:

image

Closing with hearts. 💞

PR #1079, opened 2022-06-27
#1058: add 'verbose' property

### LINKED ISSUE #1058: Add `verbose` option to log something at `info` level if task is skipped
**Is your feature request related to a problem? Please describe.**
It would be nice to have some confirmation in the logs that the task is being attempted without having to fill the logs with full debugging information.

**Describe the solution you'd like**
A new parameter:
```java
@SchedulerLock(name = "myTask", verbose = true)
```
And the potential log output:
```
[INFO] Not executing 'myTask'. It's locked until 2022-06-08T14:47:56.569Z.
```

**Describe alternatives you've considered**
Enabling debugging for `n.j.s.core.DefaultLockingTaskExecutor` and then filtering the output is a possibility, but the above solution seems more elegant.

meridbt:
Closes #1058 
Add 'verbose' property to log at INFO level the case scheduled task can't be executed due to already locked state


meridbt:
Hi, Lukas. Thank you for you fast reply. I need some time to dive into Logback project

PR #1085, opened 2022-07-01
Support ReactiveRedisLockProvider

ericwcc:
Hi team, our team uses spring boot webflux and spring-data-redis to support a highly scalable platform and didn't find any LockProvider support for ReactiveRedisConnectionFactory. It is part of the spring-data-redis. Similar to R2DBC, the only new maven dependency is `reactor-core`. May I merge this PR to improve the overall support for spring reactor project?

### DIFF
--- a/providers/redis/shedlock-provider-redis-spring/pom.xml
+++ b/providers/redis/shedlock-provider-redis-spring/pom.xml
@@ -28,6 +28,7 @@
             <groupId>io.projectreactor</groupId>
             <artifactId>reactor-core</artifactId>
             <version>3.4.19</version>
+            <optional>true</optional>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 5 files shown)


providers/redis/shedlock-provider-redis-spring/pom.xml
```diff
@@ -24,6 +24,12 @@
             <version>2.7.1</version>
         </dependency>
 
+        <dependency>
+            <groupId>io.projectreactor</groupId>
+            <artifactId>reactor-core</artifactId>
+            <version>3.4.19</version>
```
lukas-krecan, line 30: Can you please mark it as optional?

lukas-krecan (commented):
Hi, thanks a lot, I will take a look at it next week.

lukas-krecan:
Just to be sure I understand the motivation. You need the provider because you have `ReactiveRedisConnectionFactory` and you do not want to create a non-reactive one? Is that so?

ericwcc:
> Just to be sure I understand the motivation. You need the provider because you have `ReactiveRedisConnectionFactory` and you do not want to create a non-reactive one? Is that so?

yes

PR #1104, opened 2022-07-18
Adding support to new Elasticsearch Client 8

MarAra:
Please consider this PR which I'm pretending to use in my current project.


lukas-krecan:
Hi, thanks. Can you please create a new module with this code so we do not break existing clients? Just copy and paste what you have to a new module named `shedlock-provider-elasticsearch8` (or some better name).

lukas-krecan:
Released as 4.39.0

PR #1128, opened 2022-08-08
Only cache lock after a successful call to insertRecord or updateRecord

### LINKED ISSUE #753: Not executing task if create 'shedlock' table after the application was started
I am using the 4.28 version and JDBC template lock provider.

If I start the application first, then I create the table 'shedlock' in database. 
The lock will not work and keep saying "Not executing task. It's locked."  Shedlock will not insert the lock even the table is created later. Got to restart the instance to solve the problem.

It seems that the doLock() method in class StorageBasedLockProvider is doing caching in lockRecordRegistry even the insertRecord() not succeed.  
```
protected boolean doLock(LockConfiguration lockConfiguration) {
        String name = lockConfiguration.getName();
        if (!lockRecordRegistry.lockRecordRecentlyCreated(name)) {
            // create record in case it does not exist yet
            if (storageAccessor.insertRecord(lockConfiguration)) {
                lockRecordRegistry.addLockRecord(name);
                // we were able to create the record, we have the lock
                return true;
            }
            // we were not able to create the record, it already exists, let's put it to the cache so we do not try again
            lockRecordRegistry.addLockRecord(name);
        }

        // let's try to update the record, if successful, we have the lock
        return storageAccessor.updateRecord(lockConfiguration);
    }
```

basclaessen:
Only cache lock in `lockRecordRegistry` after a successful call to `insertRecord` or `updateRecord`.

Fixes #753

### DIFF
--- a/shedlock-core/src/main/java/net/javacrumbs/shedlock/support/StorageBasedLockProvider.java
+++ b/shedlock-core/src/main/java/net/javacrumbs/shedlock/support/StorageBasedLockProvider.java
@@ -83,12 +83,14 @@ public class StorageBasedLockProvider implements ExtensibleLockProvider {
                 // we were able to create the record, we have the lock
                 return true;
             }
-            // we were not able to create the record, it already exists, let's put it to the cache so we do not try again
-            lockRecordRegistry.addLockRecord(name);
         }
 
         // let's try to update the record, if successful, we have the lock
-        return storageAccessor.updateRecord(lockConfiguration);
+        if (storageAccessor.updateRecord(lockConfiguration)) {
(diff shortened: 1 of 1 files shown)


lukas-krecan:
Thanks for the PR. This makes the cache in lockRecordRegistry much less effective but it solves your edge-case. I do not think it's a good trade-off to make the code less effective for everyone to fix the code for people who start the app and then they create the DB. I will try to figure-out solution with a better trade-off.

lukas-krecan:
What about this, would it help? https://github.com/lukas-krecan/ShedLock/pull/1129

PR #1129, opened 2022-08-08
#753 Clear cache if the update throws an exception

lukas-krecan:


basclaessen:
A nice solution to remove the lock from the cache when the update fails.
Thank you for the quick fix!

PR #1136, opened 2022-08-16
couchbase collection support

mesuutt:
I made a change in couchbase provider in order to use custom collection without breaking change.

We can use a custom collection in default scope like below with the PR:

```java
new CouchbaseLockProvider(bucket.collection("lock"));
```

Or with custom scope and collection:

```java
new CouchbaseLockProvider(bucket.scope("test").collection("lock"));
```

------

Bean usage example:

```java
@Bean("lockCollection")
public Collection lockCollection(Cluster cluster) {
    return cluster.bucket(getBucketName()).collection("lock");
}

@Bean
public CouchbaseLockProvider lockProvider(@Qualifier("lockCollection") Collection collection) {
    return new CouchbaseLockProvider(collection);
}
```

PR #1137, opened 2022-08-16
feature/#1006 Support couchbase collections

daskerim:
Current version is only supporting default-collection. With this development,
CouchbaseLockProvider is supporting different Collection with-in a Bucket and
this is provided by a class which is implemented by using
CouchbaseLockConfiguration interface.

Couchbase needed a new version because of Collections support. And which is why
it's updated from 3.0.5 to 3.2.0 and has not been observed any backward
compatibility issue.

README.md updated for these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,42 @@ public CouchbaseLockProvider lockProvider(Bucket bucket) {
     return new CouchbaseLockProvider(bucket);
 }
 ```
+<hr>
+Or, to use Couchbase collections, you can create new class by implementing CouchbaseLockConfiguration
+and override methods.
+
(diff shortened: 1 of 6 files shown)


lukas-krecan:
Hi, thanks. Is your PR somehow related to https://github.com/lukas-krecan/ShedLock/pull/1136?

lukas-krecan:
Thanks for the PR, but I prefer #1136, it's much smaller. Thanks again

daskerim:
Hi, we are on the same team with the owner of the other PR and we opened two PRs without knowing each other's doing it 🙂

PR #1155, opened 2022-09-05
Update MySqlServerTimeStatementsSource.java

onlinemd:
To Avoid duplicate error on server restart

### DIFF
--- a/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/MySqlServerTimeStatementsSource.java
+++ b/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/MySqlServerTimeStatementsSource.java
@@ -35,9 +35,9 @@ class MySqlServerTimeStatementsSource extends SqlStatementsSource {
 		" ON DUPLICATE KEY UPDATE " + updateClause();
     }
 	
-	@NonNull
+    @NonNull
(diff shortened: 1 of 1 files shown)


lukas-krecan:
Hi, thanks. Can you please try to run the `MySqlJdbcTemplateLockProviderIntegrationTest`. If you do, you will find out
1. Your code does not compile
2. You need to specify the WHERE clause for the update, otherwise the lock will be always rewritten, even if it's held by another process. (see [this](https://github.com/lukas-krecan/ShedLock/blob/master/providers/jdbc/shedlock-provider-jdbc-template/src/main/java/net/javacrumbs/shedlock/provider/jdbctemplate/PostgresSqlServerTimeStatementsSource.java#L40))

Can you please fix it? (I am not sure that MySQL supports WHERE clause in insert, maybe that's the reason why it was not implemented. It's an old code, so I do not remember anymore.) Thanks again

shenghu:
@lukas-krecan due to the project limitation, we have to use 4.43.0. I understand that the exception has no harm. However, I would like to hide the exception to avoid confusion. Is it ok to backport this change to 4.x? If yes, I can help to create PR.

lukas-krecan:
Released as 4.44.0

marcelocamanho:
> Released as 4.44.0

hi, this doesnt seem to be backported in current 4.44.0 version, nor the current version. was it rolled back? thanks

lukas-krecan:
This PR was never merged as it never worked. But IGNORE keyword has been added so there is no error logged.

Issue #1284, opened 2022-12-16
SpEL referencing method parameter not working


spekdrum:
Since 0.17.0 version ShedLock supports SpEL, but this is not working:

```
@SchedulerLock(name = "{#hotelCode}", lockAtMostFor = "60s")
public void printMessage(final String hotelCode, final String message) {
...
}
```

This is what we get in Redis:
![imagen](https://user-images.githubusercontent.com/3244109/208083510-7b348817-b0b1-44e4-b300-4513d0e3e23b.png)

Also tried withoud curly braces "{ }"

lukas-krecan:
Thanks for reporting. I will look into that.

lukas-krecan:
Hi, this is surprisingly hard to implement. Have you seen any annotation other than `@Cacheable` supporting this?

ipalbeniz:
I found an example in this article, maybe it will help you understand how it could be done: https://www.springcloud.io/post/2022-03/spel-aop-log/

lukas-krecan:
Hi, I try to focus the library for the original purpose - scheduled task lock. I am afraid your use-case is out of the scope. That being said, you can use [this approach](https://github.com/lukas-krecan/ShedLock#locking-without-a-framework) which should solve your use-case.

spekdrum:
Hi Lukas, I understand your point. IMHO the `@SchedulerLock` annotation is one of the key points of this library. It would be great to comply with this SpEL use case giving a wider support. The need to partition your locks can be a common requirement for big projects, for example, partition a lock per hotel for a method instead of locking for the entire hotel chain. I know this can be done programatically, as you pointed at the documentation, but if this can be done just using an annotation and SpEL would be great.

lukas-krecan:
Hi, please correct me if I am wrong, but I do not think there is any SpEL spec of how to resolve method parameters. The only place I could find is the Cacheable annotation and the code seems to be really specific for the use-case of Caching.

Moreover, your use-case is clearly not scheduling but generic locking.

That being said, if you want to, you can submit a PR. If it does not bring too much complexity to the project, I will be happy to merge it. Otherwise, please implement it just for your project, Spring AOP is not that hard.

Yannis-01:
This capability like java synchronized(object), synchronized(object) and synchronized method have different application scenarios. synchronized(object) is fine-grained

lukas-krecan:
Released as 5.1.0

PR #1298, opened 2023-01-02
#1284 Added SpEL support to @SchedulerLock name attribute

ipalbeniz:
This PR adds SpEL support to @SchedulerLock "name" attribute (#1284)

I think it does not add much complexity to ShedLock.

What do you think @lukas-krecan?


ipalbeniz:
By the way, the SpEL parser uses a `TemplateParserContext` with the default prefix `#{` and suffix `}` in order to mix literal text with one or more SpEL evaluation blocks.

More info: https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#expressions-templating

lukas-krecan:
Looks good, thanks

Issue #1362, opened 2023-03-03
Configuration for upper- and lowercase table/column names for JDBC


Ragin-LundF:
**Problem**
Depending on the database (settings), it might be, that tables are uppercase and that the system reacts case-sensitive.
We have this "problem" at the moment with MariaDB, because we basically have all table names defined as upper case.

**Solution**
In such a case, it would be much easier to tell the JdbcProvider builder to use uppercase instead of redefining the table name and columns.

I prepared a PR, which solves the issue with a very small change.

lukas-krecan:
Released as 5.2.0

PR #1363, opened 2023-03-03
Issue #1362: Give the possibility to make table- and column names uppercase

Ragin-LundF:
Solves the issue #1362 with the possibility to define in the builder if table- and column names should be uppercase.


lukas-krecan:
Hi, thanks for the PR. I will look at it in next few days.

lukas-krecan:
Thanks, I will release it in next few days

PR #1457, opened 2023-06-17
GCP Datastore provider

mmastika:


providers/datastore/shedlock-provider-datastore/src/main/java/net/javacrumbs/shedlock/provider/datastore/DatastoreStorageAccessor.java
```diff
@@ -0,0 +1,168 @@
...
+                    var lockUntilTs = nullableTimestamp(entity, this.fields.lockUntil());
+                    return lockUntilTs != null && lockUntilTs.isBefore(now);
+                })
+                .map(entity -> {
+                    txn.put(Entity.newBuilder(entity)
+                        .set(this.fields.lockUntil(), fromInstant(until))
+                        .set(this.fields.lockedAt(), fromInstant(ClockProvider.now()))
+                        .set(this.fields.lockedBy(), this.hostname)
+                        .build());
+                    txn.commit();
```
lukas-krecan, line 89: Wouldn't it be better to commit it in `doInTxn`?

providers/datastore/shedlock-provider-datastore/src/main/java/net/javacrumbs/shedlock/provider/datastore/DatastoreLockProvider.java
```diff
@@ -0,0 +1,78 @@
...
+                return this;
+            }
+
+            public Configuration build() {
+                return new Configuration(this.entityName, this.fields, this.datastore);
+            }
+        }
+    }
+
+    public record Fields(String lockUntil, String lockedAt, String lockedBy) {
```
lukas-krecan, line 71: Maybe `FieldNames` is a better name, isn't it?

lukas-krecan (commented):
Looks great, thanks. Can you please test both concurrency modes? (I am not familiar with the Datastore, so I am not sure it will work, or if it makes sense)

mmastika:
> Looks great, thanks. Can you please test both concurrency modes? (I am not familiar with the Datastore, so I am not sure it will work, or if it makes sense)

According to [the reading](https://cloud.google.com/datastore/docs/concepts/transactions#concurrency_modes) The default concurrency mode on newly provisioned Datastore instance is PESSIMISTIC (running firestore in datastore mode) and based on the [documentation of the emulator](https://cloud.google.com/sdk/gcloud/reference/beta/emulators/datastore/start) there doesn't seem a way to set the concurrency mode on the emulator itself. 

To make matters worse, there isn't a mechanism to introspect the concurrency mode from the client that I'm aware of so as to fail fast upon finding on an unsupported concurrency mode. I have yet to test it on other concurrency mode as I am limited on privilege to do so.

The option that I can see is to have this documented in the usage guide. What do you think?

lukas-krecan:
It's OK by me. Can you please do the small changes from the comments? Then we can merge it and release it. Thanks again

providers/datastore/shedlock-provider-datastore/src/main/java/net/javacrumbs/shedlock/provider/datastore/DatastoreStorageAccessor.java
mmastika: I had a different idea in the beginning; but given the final product it makes very little sense to have it there. Will have the commit in the txn to reflect the now observed behaviour.

lukas-krecan (approved):

Issue #1505, opened 2023-08-24
Make `JedisLockProvider` implement `ExtensibleLockProvider` to allow using it with `KeepAliveLockProvider`


shotmk:
I want to use `JedisLockProvider` in non-spring application that has Jobs. The problem i see with this is that it is not possible to define the right TTL for the Lock of the Job. Setting TTL too short may cause lock to expire before Job ends. Setting TTL too high may cause issue when application will exit unexpectedly and restart, the lock in Redis will prevent the Job from running.
I believe that for the Jobs with not deterministic run times the best approach would be using KeepAlive job setting the TTL significantly lower then average job run time. 

Unfortunately `JedisLockProvider` does not implement the `ExtensibleLockProvider` and `RedisLock` does not implement the `extend()` method. 

According to the Redis documentation here: https://redis.io/commands/expire/
It looks like it is possible to change the expiration of existing keys in Redis. So i do not see the technical limitation why RedisLock do not implement the `extend()` functionality. 

WDYT? Is it doable? Does it makes sense to you? Am i missing something?

I am willing to contribute in case you agree  with me on this, and you don't see any issues my suggestion.

lukas-krecan:
Released as 5.7.0

shotmk:
@lukas-krecan can this be added as patch to version 4.xx ? I need to use it in legacy project that is running java8 with no spring, and as documentation states i need it to use 4.xx versions for that :( 

> If you are on older JDK or library, use version 4.44.0

lukas-krecan:
@shotmk Released as 4.45.0

shotmk:
@lukas-krecan just fetched the `net.javacrumbs.shedlock:shedlock-provider-redis-jedis:4.45.0` and i see that: 
> JedisLockProvider implements LockProvider
Looks like change did not reach the 4.45 version :(

lukas-krecan:
The PR should be based on `version4` branch

shotmk:
@lukas-krecan looks like this is exactly what i need, just created a PR to `version4` branch

PR #1515, opened 2023-09-05
Add ExtensibleLockProvider functionality to JedisLockProvider

shotmk:
Continuation of #1505 : Adding ExtensibleLockProvider functionality to JedisLockProvider for older Jedis version to version4


lukas-krecan:
Released as 4.46.0

shotmk:
Thanks for such a fast response and release! You are the best !

PR #1527, opened 2023-09-14
Support Micronaut 4 ConversionService type

lucjross-favor:
Hi, Lukáš,

My application uses Micronaut 4, is written in Kotlin, and uses KSP for annotation processing. The annotation SchedulerLock isn't being processed and I believe it's due to SchedulerLockInterceptor's use of ConversionService<?>. ConversionService is unparameterized in Micronaut 4 so I think KSP is quietly failing to apply the interceptor as Aspect advice on my annotated method. This hopefully is backward compatible and will get my code working.


lukas-krecan:
Hi, thanks a lot. I tried to write a test for that [here](https://github.com/lukas-krecan/ShedLock/pull/1529) but it's failing. Most likely due to invalid setup on my side. Can you please look at it?

lukas-krecan:
Hi, I am afraid the best way is to create a new micronaut4 module like [this](https://github.com/lukas-krecan/ShedLock/pull/1532). The test pases if I do it this way.

lukas-krecan:
I have created a new module. Feels safer this way. Will be releases as 5.8.0 in few minutes. Thanks

lucjross-favor:
Sorry for slow reply. This is amazing! Thank you very much!

Issue #1779, opened 2024-02-19
Spring shedlock table update row commit getting failed when the backend DB (Mysql InnoDB cluster) is running in multi-primary mode


VenkateshJhipster:
A spring boot microservice deployed in OCP and makes connection to the backend database using mysql-router image. This service contains a scheduler class (using @scheduled annotation) to do certain workflows based on a cron expression. This microservice is running with multiple nodes.

In order to stop the scheduler class getting executed from all the nodes, SHEDLOCK implementation was done. This setup was working perfectly when the backend database (Mysql Inno DB cluster) is running in Single-primary mode.

Recently, we have changed the database mode from single-primary to multi-primary mode, Similarly mysql-router configurations were also changed from first available mode to round robin mode. Post this implementation, when the SHEDLOCK library is trying to update the lock timings in SHEDLOCK table, the jdbc execution gets failed with below error. Kindly help for solving this issue.

**Exception details:**
```
ERROR Unexpected exception
org.springframework.transaction.TransactionSystemException: JDBC commit failed
    at org.springframework.jdbc.datasource.DataSourceTransactionManager.translateException(DataSourceTransactionManager.java:435) ~[spring-jdbc-6.0.9.jar!/:6.0.9]
    at org.springframework.jdbc.datasource.DataSourceTransactionManager.doCommit(DataSourceTransactionManager.java:336) ~[spring-jdbc-6.0.9.jar!/:6.0.9]
    at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:743) ~[spring-tx-6.0.9.jar!/:6.0.9]
    at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:711) ~[spring-tx-6.0.9.jar!/:6.0.9]
    at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:152) ~[spring-tx-6.0.9.jar!/:6.0.9]
    at net.javacrumbs.shedlock.provider.jdbctemplate.JdbcTemplateStorageAccessor.execute(JdbcTemplateStorageAccessor.java:119) ~[shedlock-provider-jdbc-template-5.3.0.jar!/:?]
[...]

PR #1807, opened 2024-02-29
#1779 Ability to rethrow unexpected exception in JdbcTemplateStorageAccessor

lukas-krecan:


VenkateshJhipster:
@lukas-krecan  Can pls confirm if this fix is available to download from Maven repository?

lukas-krecan:
Hi, first of all it's not a fix. It will allow you to do retry if you wrap the LockProvider by your own wrapper similar to [this](https://github.com/lukas-krecan/ShedLock?tab=readme-ov-file#customization). I have not merged it yet as I am waiting for your confirmation that it's something you want to do. I do not want to merge something that's not going to be used.

To reiterate. If we merge it, you will have to wrap LockProvider, catch the exception and retry. Is it something you want to try or not?

ivankhoosty:
Hi @lukas-krecan , this is a very useful change, I was just about to raise a PR and then noticed you already have this :)

we hit this during DB failover, ERROR: cannot execute UPDATE in a read-only transaction; nested exception is org.postgresql.util.PSQLException: ERROR: cannot execute UPDATE in a read-only transaction.  
Basically, DB goes into read-only mode, and all insert/update queries fail with UncategorizedSQLException.

we are using shedlock for scheduling, but also as standalone distribute lock provider. while returning empty lock is not catastrophic for scheduling (a run is missed, but presumably the next one will be ok), it's pretty bad for locking, as it basically means the lock cannot be obtained and we skip certain actions.

my question is why have you only singled out UncategorizedSQLException and not the whole catch block (regardless of exceptions listed, they are basically classed by shedlock as "Unexpected exceptionS", maybe the change would be even more useful if LockConfiguration had throwUnexpectedExceptionS (plural), and then all exceptions in that catch block would qualify .

lukas-krecan:
Cool, I will merge it then

lukas-krecan:
@ ivankhoosty Released in 5.13.0

lukas-krecan:
@ivankhoosty Released in 5.13.0

ivankhoosty:
Thanks for a quick turnaround on this !

PR #2045, opened 2024-07-23
feat(spring-redis): added lock extension for spring redis lockprovider

shubhajyoti-bagchi-groww:
Adding lock extend support to spring redis lock provider


lukas-krecan:
Thanks again

lukas-krecan:
Released as 4.48.0

Issue #2153, opened 2024-09-22
Add a ShedLock provider for commercetools


studix:
**Is your feature request related to a problem? Please describe.**
We are using the commercetools platform to build our e-commerce solutions. If you use this platform the backend applications usually have no database other than commercetools itself.

**Describe the solution you'd like**
We'd like to use a provider that uses commercetools custom objects to persist lock information. Custom objects are a key/value document store that is part of the API extensibility provided by the platform.

**Describe alternatives you've considered**
We considered using the existing Redis providers. This would require us to setup an additional infrastructure component which leads to increased infrastructure costs.

**Additional context**
The provider would use the HTTP AP provided by commercetools: [https://docs.commercetools.com/api](https://docs.commercetools.com/api)

studix:
We can work on that and we would provide a PR soon at [https://github.com/BisonSchweizAG](https://github.com/BisonSchweizAG)

lukas-krecan:
Hi, please send a PR. The main condition is that you also provide passing tests (extending AbstractLockProviderIntegrationTest). Thanks a lot.

studix:
I couldn't get the AbstractLockProviderIntegrationTest green in the build pipeline. One of the tests (FuzzTester) performs a lot of concurrent requests which seems to be an issue with the commercetools api mock that is used in my integration test.

I'm closing this issue.

lukas-krecan:
Thanks a lot for trying.

PR #2154, opened 2024-09-22
#2153 Add a ShedLock provider for commercetools

studix:
ShedLock provider that uses commercetools custom objects to persist lock information. Custom objects are a key/value document store that is part of the API extensibility provided by the platform.

Could not inherit the integration test from AbstractLockProviderIntegrationTest as there is no test container for commercetools. Used a http mock server instead.


lukas-krecan:
Hi, thanks. But I am afraid I can't merge it without the integration tests as I will not be able to support it. The integration tests give me confidence that the lock provider works as expected, mocked tests do not help much in this regard. We either have to somehow figure-out how to make the tests work, or you can release your Lock Provider as a standalone project and I will hapilly link it from the documentation.

studix:
I found a docker image that mocks the whole commercetools server. With this I could pass the integration tests.

studix:
Looks like the test fuzzTestShouldPass is flaky. What is this test doing?

lukas-krecan:
It's the most interesting test. It's executing [FuzzTester](https://github.com/lukas-krecan/ShedLock/blob/master/shedlock-test-support/src/main/java/net/javacrumbs/shedlock/test/support/FuzzTester.java) which is incrementing a counter from multiple threads and checkig if the resulting number has the expected value. If the locking does not work, this test usually fails. It's unfortunately hard to tell why the test failed.

studix:
I got the FuzzTester green with the mocked and the real commercetools API. It's stable on my machine. But it still fails on the pipeline.

I'm giving up and I'm closing the pull request. You can also close #2153 if you want.

lukas-krecan:
Hi, I looked into the code and there may be a conflict on Insert. Imagine all the threads are executed in parallel. All threds call find [here](https://github.com/BisonSchweizAG/ShedLock/blob/feature/commercetoolsLockProvider/providers/commercetools/shedlock-provider-commercetools/src/main/java/net/javacrumbs/shedlock/provider/commercetools/CommercetoolsStorageAccessor.java#L29), they got null as there is no record and then the call to `createOrUpdateLockValue` succeed as no `version` is sent. I am not familiar with the API, so I am not sure it's the case. What do you think?

studix:
Calling find would actually not be necessary as the api has a createOrUpdate semantic. It is guaranteed by commercetools that there will be no duplicate custom obects with the same key (lock name). I‘m catching this in the insert method and return empty

Issue #2217, opened 2024-10-31
Support for multiple, different LockProviders (i.e. in a modular monolith - each module providing its own locks storage)


risto-uibo-inbank:
Currently 

- `@SchedulerLock` behind the scenes requires a single, application-wide `LockProvider` bean
- In other words - it's not possible to specify, what locks storage (what kind of lock provider) to use for different scheduled jobs, when applying `@SchedulerLock` annotation on them.
- The annotation processing logic behind the annotation always expects exactly one, single `LockProvider` bean from the Spring context

Issue

- This creates a runtime decoupling problem in modular monoliths. Especially, in projects applying vertical-slice architecture. 
- In this architecture, the application is composed of independent modules - along the logical bounded contexts (BC-s) of the system.
- Each BC  "module" is either a completely separate gradle/maven module, or a distinct root package of the application.
- Each BC module contains the application layer & application config (including Spring beans), core (domain) layer and data providers / adapters (repositories) layer. 
- So each module is as independent as possible - both in terms of its code dependencies as well as runtime dependencies (application context).
- Since currently only a single `LockProvider` bean can (and must) be configured, it must be a shared one between the modules - an application-wide bean. While, each module usually wants to have its own DB schema where it needs to store locks, or use a completely different locking mechanism than other modules, for various reasons. Currently that's not possible.

Solution

- Extend the `@SchedulerLock` annotation to support specifying its `LockProvider`, by a `lockProvider=` qualifier property, which would be the name of the `LockProvider` bean that is to be used here.
- In the annotation processing, the qualified `LockProvider` bean would be requested from Spring Application Context - and the LockManager which executes the scheduled code, is created with this specific LockProvider.

Alternatives
- As in my work, we've come across this need, we have created a sort of a parallel annotation to the `@SchedulerLock` annotation - `@QualifiedSchedulerLock`, which does all the same things, except it also requires specifying the `LockProvider`:

```
@Target({METHOD, ANNOTATION_TYPE})
@Retention(RUNTIME)
public @interface QualifiedSchedulerLock {

  String lockProvider();

  String name();

  String lockAtMostFor() default "";

  String lockAtLeastFor() default "";
}
```

- And to go along with that annotation, a processing aspect, where the main logic snippet is:
```
@Around("@annotation(com.our.company.....schedulerlock.QualifiedSchedulerLock)")
  public Object acquireQualifiedSchedulerLock(ProceedingJoinPoint scheduledMethod) {
[...]

PR #2234, opened 2024-11-06
Preparation for v6 release

lukas-krecan:


szpak:
@lukas-krecan Is there any migration needed for project using 5.16 with just one lock provider? I wasn't able to find the answer in the release notes, nor the original issue #2217 (which seems to be still one anyway).

lukas-krecan:
No, if you have just one lock provider, everything should work as before. Please let me know if it's not the case. I will clarify the release notes

PR #2257, opened 2024-11-18
NATS JetStream support

Saphri:
Adding support for using NATS JetSteam key-value stores as lock backend

### DIFF
--- a/providers/mongo/shedlock-provider-mongo-reactivestreams/src/test/java/net/javacrumbs/shedlock/provider/mongo/reactivestreams/ReactiveStreamsMongoLockProviderIntegrationTest.java
+++ b/providers/mongo/shedlock-provider-mongo-reactivestreams/src/test/java/net/javacrumbs/shedlock/provider/mongo/reactivestreams/ReactiveStreamsMongoLockProviderIntegrationTest.java
@@ -45,7 +45,7 @@ public class ReactiveStreamsMongoLockProviderIntegrationTest extends AbstractExt
     private static final String DB_NAME = "db";
 
     @Container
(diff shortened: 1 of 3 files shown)


lukas-krecan:
Thanks a lot, I will look at it soon

lukas-krecan:
Hi, thanks again. It seems to me, that there is another limitation, lockAtLeastFor/Until for acutally does not work, if the task takes shorter than  lockAtLeastFor, the lock is held for lockAtMostFor, do I get it right? If that's so, let's log a warning if lockAtleast for is nonzero and mention it in the documentation. Any idea when NATS 2.11 should be released?

Saphri:
Yes you are correct.. I will update the documentation and add the warning 

I did start on putting data into the key, but then started to worry about race conditions if I had to first read and then act on that. Here we dont use the atLeastFor feature at all, and only use this for ensuring some physical devices only receives commands from one process at a time

NATS 2.11 should be close,.. still claims to be coming in 2024

lukas-krecan:
Thanks a lot, how urgently do you need it? Does it make sense to wait few weeks for 2.11?

lukas-krecan:
Hi, I appreciate it. I prefer to keep the PR open and wait for the 2.11. The reason is that if they indeed deliver more fine-grained TTLs, the move from the current solution to the new one would be backwards incompatible, so I would have to keep both versions for some time. Thanks for understanding. Can you please ping me if there are any updates?

lukas-krecan:
Hi, it seems 2.11 is out. Do you want to revive the PR?

Saphri:
Yes I'll take a look at it

PR #2290, opened 2024-12-06
Make Spring integration with InterceptMode.PROXY_METHOD work for Spring versions prior to 5.0.0

ellebrecht:
When using ShedLock 4.x in a Spring Boot 1.5/Spring 4.3 application with InterceptMode.PROXY_METHOD, this fails with `NoSuchMethodError` due to usage of an `AnnotationMatchingPointcut` constructor that was added in Spring 5.0.0. 

It can easily be made to work by checking the Spring major version and using the appropriate constructor. With this fix, Spring 4.x-based applications should work again.

### DIFF
--- a/spring/shedlock-spring/src/main/java/net/javacrumbs/shedlock/spring/aop/MethodProxyScheduledLockAdvisor.java
+++ b/spring/shedlock-spring/src/main/java/net/javacrumbs/shedlock/spring/aop/MethodProxyScheduledLockAdvisor.java
@@ -44,11 +44,33 @@ class MethodProxyScheduledLockAdvisor extends AbstractPointcutAdvisor {
 
     @NonNull
     private static AnnotationMatchingPointcut methodPointcutFor(Class<? extends Annotation> methodAnnotationType) {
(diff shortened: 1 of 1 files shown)


lukas-krecan:
Hi, thanks for your contribution. I don't see why to support Spring 4.3. It's really old and unsupported. Maybe an old version of ShedLock will work with it.

ellebrecht:
I realize that Spring 4 went EOL in 2020, however I am currently tasked with maintaining a legacy application stuck at said version. Sadly, I know of a significant number of productive applications in large corporations in the same situation.

If you don't want to support it, I suggest to document version compatibilities in more detail.
Currently, the README only says: _If you are on older JDK or library, use version 4.44.0_
It would be helpful and time-saving to learn early that ShedLock does not plan to support Spring versions prior to 5.

lukas-krecan:
Hi, I have just added the [compatibility matrix](https://github.com/lukas-krecan/ShedLock?tab=readme-ov-file#compatibility-matrix)

PR #2335, opened 2025-01-07
Opensearch java LockProvider

lukas-krecan:


README.md
```diff
@@ -599,6 +599,13 @@ import static net.javacrumbs.shedlock.provider.opensearch.OpenSearchLockProvider
 
 ...
 
+@Bean
```
lukas-krecan, line 602: TODO: Fix this

providers/opensearch/shedlock-provider-opensearch-java/src/main/resources/shedlock.settings.json
```diff
@@ -0,0 +1,6 @@
+{
```
lukas-krecan, line 1: @harisonde Do you know if this file is needed?

harisonde: I do not believe so. This file along with `shedlock.mapping.json` file is not used at all. Do you know if this file `shedlock.mapping.json` is indeed used anywhere? and required unless to indicate the structure of the index.

PR #2490, opened 2025-03-28
Add shedlock-provider-jdbc to bom

svenallers:
Hi,
we realized that shedlock-provider-jdbc was missing in the bom. I hope it's fine that I just opened a PR without creating an issue first as this is a very small and trivial change.


lukas-krecan:
Thanks a lot

Issue #2405, opened 2025-02-18
Feature Enhancement Proposal: Support for Sort Keys for DynamoDBAsyncClient in Shedlock


avanisH8715:
Hello **@lukas-krecan** 

Our team is using Shedlock for task scheduling in our project, and we are using DynamoDB as our database. We have identified areas where the Shedlock DynamoDB provider could be improved:

**Inclusion of Sort Key**: Currently, to add sort key in DynamoDB provider is missing. This limitation poses a challenge for us as we have existing tables in production which contains both partition key and sort key and whenever Shedlock try to write to DynamoDB we'll get DynamoDB exception(Status Code: 400) due to schema mismatch as our table expects a sort key but Shedlock is providing empty/null value to it. It would be helpful if the sort key could be added when initializing the DynamoDB provider. So if a table contains sort key, it will initialize it otherwise will ignore this parameter. This flexibility would allow us to utilize our existing tables without having to modify them.

If you need any support or assistance in implementing this feature, I would be happy to contribute to the development effort.

Thanks!

avanisH8715:
Hi Lucas, Sorry for late reply.
I'm working on the changes, will share pull request with you soon.

lukas-krecan:
No problem, thanks

PR #2530, opened 2025-04-24
To add support for Sort Keys for DynamoDBAsyncClient in Shedlock

avanish-p1:
- https://github.com/lukas-krecan/ShedLock/issues/2405

Changes:

- Added new constructor to support sortkey value argument. To maintain compatibility with the previous constructor, I have assigned sortkey's default value to null.
- If sortkey name is not provided, DynamoDB key will be Partition key else key will contain both Partition Key and Sort Key.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <jdk.version>17</jdk.version>
         <spring-boot-current.version>3.4.2</spring-boot-current.version>
-        <junit.ver>5.12.1</junit.ver>
+        <junit.ver>5.12.2</junit.ver>
         <test-containers.ver>1.20.6</test-containers.ver>
         <kotlin.ver>2.1.20</kotlin.ver>
         <micronaut4.version>4.8.0</micronaut4.version>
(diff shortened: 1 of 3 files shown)


lukas-krecan:
@avanish-p1 Hi, thanks a lot for the PR. Can you please explain the use case a bit more? In the issue you write

> we have existing tables in production which contains both the partition key and sort key

How come you have a sort key for ShedLock table? My expectation is that you would create a new table for ShedLock, but I am not familiar with DynamoDb so I am most likely missing something.

avanish-p1:
> @avanish-p1 Hi, thanks a lot for the PR. Can you please explain the use case a bit more? In the issue you write
> 
> > we have existing tables in production which contains both the partition key and sort key
> 
> How come you have a sort key for ShedLock table? My expectation is that you would create a new table for ShedLock, but I am not familiar with DynamoDb so I am most likely missing something.

Hello @lukas-krecan,

Thank you for your response and for clarifying your expectations regarding the use of DynamoDB with Shedlock. You are correct that creating a dedicated table for Shedlock is a widely adopted approach, especially when designing systems with specific requirements in mind. However, in our case, we follow organizational best practices that encourage the consolidation of tasks within a single DynamoDB table rather than creating separate tables for each use case.

While creating a separate table for Shedlock would resolve the issue at hand, it introduces additional infrastructure management overhead. This approach may not align with our current operational practices, as it requires maintaining multiple tables, which can complicate administration and monitoring.

By incorporating support for sort keys in the DynamoDB provider, we can ensure that Shedlock remains flexible and adaptable to various use cases. This enhancement would allow users who prefer to utilize a single table with both partition and sort keys to seamlessly integrate Shedlock into their existing infrastructure. At the same time, it would not impose any limitations on users who opt to create a dedicated table for Shedlock.

In summary, this change would provide greater flexibility and compatibility, enabling all users—whether they choose to use a single table or a separate table—to fully leverage the benefits of the Shedlock library.

Please let me know if you need further clarification on this feature. I’m happy to collaborate and ensure this enhancement aligns with the library’s design principle

PR #2552, opened 2025-05-01
Add support for Redis lettuce

s4got10dev:
Hi @lukas-krecan 

I want to use ShedLock within my Micronaut application that uses Redis, but I would like to avoid having both Lettuce and Jedis connections.
Can you, please check implementation.

Thanks a lot


lukas-krecan:
Thanks a lot, merged as part of #2556

Issue #2583, opened 2025-05-14
Redis Lock implementation do not account for race-conditions


shotmk:
Current implementation of RedisLock do not account for race-conditions. 
Leaving a possibility of one thread / process unlocking a Lock acquired by another thread / process.
Scenario: 
1. Thread A takes a lock with 30s TTL 
2. Thread A lags in processing the request and it finished working on the task in 31s (lock already expired in Redis)
3. Thread B successfully acquired the lock (it was expired in Redis), and almost immediately thread A will release the lock (it finished handling previous request)

Redis posted an article on how to properly implement the distributed lock: 
https://redis.io/docs/latest/develop/use/patterns/distributed-locks/

The important part of the article is this: 

To acquire the lock, the way to go is the following:

    SET resource_name my_random_value NX PX 30000
The command will set the key only if it does not already exist (NX option), with an expire of 30000 milliseconds (PX option). The key is set to a value “my_random_value”. This value must be unique across all clients and all lock requests.

Basically the random value is used in order to release the lock in a safe way, with a script that tells Redis: remove the key only if it exists and the value stored at the key is exactly the one I expect to be. This is accomplished by the following Lua script:

```
if redis.call("get",KEYS[1]) == ARGV[1] then
    return redis.call("del",KEYS[1])
else
    return 0
end
```

This is important in order to avoid removing a lock that was created by another client. For example a client may acquire the lock, get blocked performing some operation for longer than the lock validity time (the time at which the key will expire), and later remove the lock, that was already acquired by some other client. Using just [DEL](https://redis.io/docs/latest/commands/del/) is not safe as a client may remove another client's lock. With the above script instead every lock is “signed” with a random string, so the lock will be removed only if it is still the one that was set by the client trying to remove it.

The code of the lib already sets a value for the key when acquiring or extending the lock. 
Need to do a slight change in code  to have truly UUID value for the lock (now it's HOST + clockTime.Milliseconds), also need to store the value in the RedisLock instance and extend / unlock only in case value in Redis equal to value stored inside lock object in java.

PR #2584, opened 2025-05-14
Fix for #2583

shotmk:
Implement Redis distributed locking respecting unique lock value to avoid removing a lock that was created by another client. (https://redis.io/docs/latest/develop/use/patterns/distributed-locks/)

### DIFF
--- a/providers/redis/shedlock-provider-redis-jedis4/pom.xml
+++ b/providers/redis/shedlock-provider-redis-jedis4/pom.xml
@@ -24,12 +24,6 @@
             <version>5.2.0</version>
         </dependency>
 
-        <dependency>
-            <groupId>net.javacrumbs.shedlock</groupId>
-            <artifactId>shedlock-test-support</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
-        </dependency>
 
         <dependency>
             <groupId>net.javacrumbs.shedlock</groupId>
--- a/providers/redis/shedlock-provider-redis-jedis4/src/main/java/net/javacrumbs/shedlock/provider/redis/jedis4/JedisLockProvider.java
+++ b/providers/redis/shedlock-provider-redis-jedis4/src/main/java/net/javacrumbs/shedlock/provider/redis/jedis4/JedisLockProvider.java
@@ -19,7 +19,9 @@ import static redis.clients.jedis.params.SetParams.setParams;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 import net.javacrumbs.shedlock.core.AbstractSimpleLock;
 import net.javacrumbs.shedlock.core.ClockProvider;
 import net.javacrumbs.shedlock.core.ExtensibleLockProvider;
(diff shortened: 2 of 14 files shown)


lukas-krecan:
Hi, thanks a lot. I will check it in the next few days.

shotmk:
By the way, I believe this approach should be adopted across all providers, irrespective of the underlying persistence storage.

lukas-krecan:
Hi, thanks. It looks great. I am just wondering what will happen if somebody who is using older ShedLock version will update to the new version? I think it should be OK as the change is mainly in the update and delete part which is only relevant for the process that keeps the lock. Do I get it right?

lukas-krecan:
One more thing, is the Lua script always enabled?

shotmk:
> Hi, thanks. It looks great. I am just wondering what will happen if somebody who is using an older ShedLock version will update to the new version? I think it should be OK as the change is mainly in the update and delete part which is only relevant for the process that keeps the lock. Do I get it right?

I believe we do not create a degradation, or problematic scenario during upgrade. 
The change is in the way we `unlock` or `extend` the lock. 
I wouldn't expect the code of the app to try to unlock the lock that was not taken by the app (during crash / restart scenario), and the app would rely on the TTL of the lock during acquire step: 
1. App takes the lock
2. App crashes 
3. App restarts 
4. App tries to acquire lock and fails (lock still in Redis until TTL) 



> One more thing, is the Lua script always enabled?

Lua scripts are enabled by default since Redis 2.6. 
There is a way to disable it with some tricks...
Rename the `eval` with this change in `redis.conf`
```
rename-command EVAL ""
```

Or in Redis 6+ disallow EVAL permission for the user: 
```
ACL SETUSER limiteduser -@all +get +set  .... -eval 
```

So maybe this needs to be documented, i come up with this using simple GPT prompt: 

Redis Lua Script Requirement (since version X.X.X)
Starting from Shedlock version X.X.X, operations such as unlocking and extending locks in Redis now use Lua scripts to ensure atomic execution and improved reliability. Lua scripting is a core Redis feature and is enabled by default in most Redis deployments.

However, please note that:

Lua scripting may be disabled or restricted in environments with strict Redis Access Control Lists (ACLs) or custom configurations.

If you are using a Redis deployment where scripting is disabled (e.g., certain managed services or locked-down enterprise setups), Shedlock's Redis integration will not function correctly unless scripting permissions are granted.

Required Permission:
Ensure the Redis user used by Shedlock has permission to execute the EVAL command.

You can verify this by checking your Redis ACL configuration:

```bash
ACL GETUSER your-shedlock-user
```

Look for the presence of the +eval command or +@all category.

For environments where scripting cannot be enabled, consider using an alternative Shedlock storage provider.

lukas-krecan:
The trade-off is pretty clear to me. On one hand the PR partially fixes a race-condition. It will still result into two tasks running, if the lock at most for is wrongly configured, but unlocking would not unlock a lock held by somebody else. Utility of the fix is pretty small as it does not solve the underlying issue - `lock at most for` is incorrectly configured. On the other hand, the fix can break perfectly valid configurations with ACLs. That being said, I will most likely make it configurable and opt-in. I already have a refactoring in https://github.com/lukas-krecan/ShedLock/pull/2593 that will allow me to do it in one place.

Issue #1254, opened 2022-11-22
Job not fired again after connection to database was lost


buschluschni:
Version: 4.42.0
Provider: JdbcTemplateLockProvider
Config: @EnableSchedulerLock(defaultLockAtMostFor = "10m", defaultLockAtLeastFor = "10m")

When connection to database is lost during a job run, it will not be triggered again, although the lock_until is in the past. 
After a restart (Spring Boot) it is triggered again as expected.

lukas-krecan:
Hi, thanks for feedback. Is there anything you see in the log's? Stack traces?

buschluschni:
No, I turned on DEBUG loglevel, there are no errors/stacktraces, just the debug message
_n.j.s.core.DefaultLockingTaskExecutor Not executing 'xxxx'. It's locked.
is written each time the job should be triggered.

buschluschni:
It only happens when connection to database is temporarily unavailable during a job run. No matter how often the job was already executed.

lukas-krecan:
From the code it can't happen 🙂 I can release a new version with more logging, would you be willing to help me to debug it?

buschluschni:
I'd like to help you debug it 👍

lukas-krecan:
Hi, I have just release version 4.43.0 with more logging, can you please try? Thank you

buschluschni:
Hi, thanks for the new version. I tried it out, but I'm afraid the output won't help you much (Shedlock lock time is 1m, Scheduler runs every min):

```
2022-12-05 06:53:00.311 DEBUG 2628 --- [   scheduling-1] n.j.s.core.DefaultLockingTaskExecutor : Locked 'PriceInitJob', lock will be held at most until 2022-12-05T05:54:00.014Z
...
2022-12-05 06:55:00.004 DEBUG 2628 --- [   scheduling-1] n.j.s.core.DefaultLockingTaskExecutor : Not executing 'PriceInitJob'. It's locked.
...
2022-12-05 06:56:00.020 DEBUG 2628 --- [   scheduling-1] n.j.s.core.DefaultLockingTaskExecutor : Not executing 'PriceInitJob'. It's locked.
```

lukas-krecan:
Can you please check what's in the ShedLock DB table (lock_until)?

lukas-krecan:
Ok, the one hour difference is an interesting clue (unless it's a coincidence). Do you have `usingDbTime` in you lock provider config or not?

[...]

PR #2807, opened 2025-08-21
#1254 Do not add record to a cache in case of a connection error

lukas-krecan:


lukas-krecan:
This would cause insert to be attempted over and over again

PR #2912, opened 2025-10-04
enforce majority read/write concerns in MongoLockProvider

Danden1:
This PR enforces ReadConcern.MAJORITY and WriteConcern.MAJORITY when acquiring a lock in MongoLockProvider.

## Motivation

In a MongoDB replica set, there is a potential issue:

A lock document can be written to the primary.

Before this change is replicated to the secondary, another scheduler instance might read from the secondary and incorrectly acquire the same lock.

## Impact

Pros: Stronger consistency, reduces risk of duplicate execution in replica set environments.

Cons: Slightly higher latency in lock operations.

Note: While this improves safety, it still does not provide a 100% guarantee of consistency. Certain edge cases (e.g., network partitions, failovers, or unexpected replication delays) may still cause rare anomalies.


lukas-krecan:
Hi thanks

lukas-krecan:
Hi, actually, I changed my mind. Everybody can configure it on the Mongo collection they pass to the LockProvider. I will enhance the documentation, to make it recommended.

PR #3033, opened 2025-11-19
Improve nullable usage

AndreiPurcaru:
Hey there Lukas!

I was updating to the latest version of ShedLock in one of my projects and noticed ErrorProne warning about the usage of the jspecfiy `@Nullable` in one of our custom implementations. The annotation is targeting `TYPE_USE` so it needs to be placed right before the type rather than on methods. There is some more info about it [here](https://github.com/jspecify/jspecify/wiki/type-usages). I took the liberty of opening this PR to change the placement so it aligns with the recommended placement. Do let me know if you have a different take on this or if there is a reason for these to be used differently.

### DIFF
--- a/cdi/shedlock-cdi/src/main/java/net/javacrumbs/shedlock/cdi/internal/SchedulerLockInterceptor.java
+++ b/cdi/shedlock-cdi/src/main/java/net/javacrumbs/shedlock/cdi/internal/SchedulerLockInterceptor.java
@@ -40,7 +40,8 @@ public class SchedulerLockInterceptor {
     }
 
     @AroundInvoke
-    @Nullable Object lock(InvocationContext context) throws Throwable {
+    @Nullable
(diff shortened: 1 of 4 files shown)


lukas-krecan:
Hi, thanks a lot, I have done it without properly reading the doc (shame on me). Error prone is failing on [this](https://github.com/lukas-krecan/ShedLock/blob/062cfb9b160f48a3f29170bbc34a50b15653e44e/spring/shedlock-spring/src/main/java/net/javacrumbs/shedlock/spring/aop/SpringLockConfigurationExtractor.java#L259) so I guess that for arrays it works differently then you expect.

lukas-krecan:
I have fixed it here https://github.com/lukas-krecan/ShedLock/pull/3038

lukas-krecan:
Merged as part of #3038

Issue #3082, opened 2025-12-09
Support R2dbcLockProvider in WebFlux controllers


Danden1:
**Is your feature request related to a problem? Please describe.**

I am using Spring WebFlux. Although I mostly rely on a scheduler, there are times when I need to trigger a task directly through an API request.

When using the R2dbcLockProvider, the scheduler works correctly, but calling the same logic through an API request results in an error.

**Describe the solution you'd like**

I would like the lock to work properly both in the scheduler and when invoked through a WebFlux controller.

**Describe alternatives you've considered**

I am planning to open a PR for this issue, so I would appreciate it if you could review it.

**Additional context**

Example:

```java
@RestController
public class HelloController {

    private final LockProvider lockProvider;
    Duration lockAtMostFor = Duration.ofSeconds(60);
    Duration lockAtLeastFor = Duration.ZERO;

    public HelloController(LockProvider lockProvider) {
        this.lockProvider = lockProvider;
    }

    @GetMapping("/test")
    @Scheduled(fixedDelay = 1000)
    public Mono<String> test() throws Throwable {
        Runnable runnable = () -> {
            System.out.println("Lock acquired, executing task");
        };
        LockingTaskExecutor executor = new DefaultLockingTaskExecutor(lockProvider);
        Instant createdAt = Instant.now();
        executor.executeWithLock(runnable, new LockConfiguration(createdAt, "lockName", lockAtMostFor, lockAtLeastFor));


        return  Mono.just("Task executed");
    }
    
}
```

```java
@Configuration
public class R2dbcConfig {

    @Bean
    public LockProvider getLockProvider(ConnectionFactory connectionFactory) {
        return new R2dbcLockProvider(connectionFactory);
    }
}
```

When I send a request to localhost:8080/test, the following error occurs.

```
Caused by: java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread reactor-http-nio-2
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:121) ~[reactor-core-3.8.0.jar:3.8.0]
	at reactor.core.publisher.Mono.block(Mono.java:1800) ~[reactor-core-3.8.0.jar:3.8.0]
	at net.javacrumbs.shedlock.provider.r2dbc.R2dbcStorageAccessor.block(R2dbcStorageAccessor.java:81) ~[shedlock-provider-r2dbc-7.2.1.jar:na]
[...]

PR #3083, opened 2025-12-09
Enable R2dbcLockProvider to Work in WebFlux Controllers

Danden1:
This PR enables R2dbcLockProvider to work properly when invoked from a WebFlux controller.

#3082 

After applying the changes in this PR, the lock works correctly when triggered through a WebFlux controller as well as through a scheduler.

I have tested this locally as well, and it works correctly without any issues.


lukas-krecan:
Hi, thanks a lot. Can you please:

1. Apply this [fix](https://github.com/lukas-krecan/ShedLock/commit/c0c15b899fc1f294c572f5f139671ceca6c856d5) so the build passes
2. Document the change in the code to explain why it's there and how it works.
3. Explain why we need both the first and the last argument in `mono.subscribe`

Danden1:
Hi! I am making good use of this open-source project.

I have applied the requested changes.

In particular, regarding the third point, I added the following clarification:

Although this is already described in the comment, when Mono.empty() is passed as an argument, onNext is never invoked, which means the consumer is not executed and no value can be retrieved. 

To handle this scenario, we explicitly call the completeConsumer, ensuring that if the consumer is not triggered, the method returns null.

Issue #3682, opened 2026-08-19
JooqLockProvider computes lock_until from a stale transaction snapshot when the DSLContext shares an ambient transaction(follow-up to #3577)


stlahxm:
Related to #3577. That issue is still waiting on one fact: whether the reporter's transaction starts well before the annotated method runs. I can't answer that for them, but I put together a standalone, deterministic reproduction that isolates the mechanism their RCA describes from their specific application. Posting it separately since it doesn't depend on their confirmation and the original thread has been quiet since 2026-08-15.

**Read the documentation**

Checked the troubleshooting section and the jOOQ caveat already in the README. This report is about a corner case that caveat doesn't fully capture - see "Root cause and prior art in this repo" below.

**Describe the bug**

1. Version: reproduced against current `main` (jOOQ dependency at 3.21.7); the affected code path is unchanged in the latest release, 7.9.0.
2. Lock Provider: `JooqLockProvider`.
3. ShedLock configuration: `DSLContext` bound to a connection that shares an already-open (auto-commit=false) transaction - e.g. jOOQ integrated with Spring's transaction manager via a transaction-aware connection provider, the same setup as the reporter in #3577.
4. ShedLock logs: none - this fails silently, no error or warning is logged anywhere.

**Expected behavior**

`lock_until` should be computed from the actual current time at the moment the lock is acquired (`now() + lockAtMostFor`), regardless of whatever transaction happens to already be open on the connection.

**Actual behavior**

`lock_until` is computed from the snapshot time of whatever transaction was already open on the connection, not the actual current time. With a short enough `lockAtMostFor` and a long enough pre-existing transaction, this can make `lock_until` already be in the past by the time it's written, letting a second node steal the lock. Reproductions below.

### Summary

`JooqStorageAccessor` computes `lock_until` using jOOQ's `currentLocalDateTime()`, which renders to `CURRENT_TIMESTAMP` on the target database. On PostgreSQL (and other databases with MVCC-style snapshot semantics), `CURRENT_TIMESTAMP` is fixed at the *start of the transaction*, not at statement execution time. If the `DSLContext` passed to `JooqLockProvider` is bound to a connection that already has a transaction open (auto-commit=false, opened before the lock is acquired), `lock_until` ends up computed relative to that old snapshot instead of the actual current time.

[...]

Issue #3577, opened 2026-07-07
JooqLockProvider (Postgres) allows two instances to acquire the same lock within the same second — both log Locked '<name>' and execute the task


manoranjanotta:
We are facing a issue on prod where Two application instances managed to acquire the same ShedLock lock  within ~274 ms of each other, both logged Locked '<name>' ... at DEBUG from DefaultLockingTaskExecutor, and both executed the guarded task. This defeats the whole purpose of ShedLock in a multi-instance deployment.
We use JooqLockProvider on Postgres. usingDbTime() is not applicable to the JOOQ provider (JDBC provider option), but the same effect should come from DB-side now() inside the SQL that JOOQ provider issues.

Item | Value
-- | --
ShedLock version | 5.13.0
Provider | net.javacrumbs.shedlock:shedlock-provider-jooq:5.13.0
Spring integration | net.javacrumbs.shedlock:shedlock-spring:5.13.0
Spring Boot | 3.x (matching Boot for Spring 6)
Kotlin | 2.2.20
JOOQ | 3.20.7
Database | PostgreSQL (driver org.postgresql:postgresql:42.7.7)
DB isolation | Default (READ COMMITTED) — Hikari default, not overridden
Deployment | 2 pods (typical), both hitting the same Postgres

Configuration:
@Configuration
@EnableScheduling
@EnableSchedulerLock(
    defaultLockAtMostFor = "PT5M",
    defaultLockAtLeastFor = "PT1M",
)
class ShedlockConfiguration {
    @Bean
    fun lockProvider(dslContext: DSLContext): LockProvider = JooqLockProvider(dslContext)
}

Observed logs (production, redacted)

2026-07-01T00:00:00.063  DEBUG  DefaultLockingTaskExecutor  Locked 'XXXXXX', lock will be held at most until 2026-06-30T18:35:00Z
2026-07-01T00:00:00.337  DEBUG  DefaultLockingTaskExecutor  Locked 'XXXXXX, lock will be held at most until 2026-06-30T18:35:00.236Z
2026-07-01T00:00:00.816  DEBUG  DefaultLockingTaskExecutor  Task finished, lock 'XXXXXX' will be released at 2026-06-30T18:31:00Z
2026-07-01T00:00:00.848  DEBUG  DefaultLockingTaskExecutor  Task finished, lock 'XXXXXX' will be released at 2026-06-30T18:31:00.236Z

Notes on the timing:
[...]

PR #3684, opened 2026-08-19
Add optional DataSource-based isolation to JooqLockProvider (fixes #3683)

### LINKED ISSUE #3683: JooqLockProvider(DSLContext) silently commits unrelated caller data when the DSLContext is bound to a connection with an already-open transaction
Related: #3577 (fixed in 7.9.0 via `NewTransactionRunner`) and #3682 (closed - covered a timestamp-computation symptom of a related but different scenario). This is about a separate, more serious side effect that neither addresses.

**Read the documentation**

Checked the troubleshooting section and the jOOQ caveat already in the README - neither covers this. Already on the latest release (7.9.0); the affected code path is unchanged there.

**Describe the bug**

1. Version: reproduced against current `main` and against the latest release, 7.9.0 (jOOQ dependency 3.21.7) - the affected code path is the same in both.
2. Lock Provider: `JooqLockProvider`.
3. ShedLock configuration: `DSLContext` bound to a connection that already has an open (`auto-commit=false`) transaction at the time a lock operation runs - e.g. `DSL.using(connection, dialect)` on a connection the caller is already using for something else.
4. ShedLock logs: none - this fails completely silently, no error or warning is logged anywhere.

**Expected behavior**

A lock operation (`insertRecord`/`updateRecord`/`extend`/`unlock`) should never affect any transaction state the caller itself is managing on that connection, regardless of whatever transaction happens to already be open on it and regardless of whether the lock operation itself succeeds.

**Actual behavior**

A single lock operation unconditionally commits the connection's current transaction - including any of the caller's own unrelated, not-yet-committed work - whether or not the lock attempt succeeds.

**Root cause**

`JooqStorageAccessor` runs lock operations through `dslContext.transactionResult(txCallable)`, which goes through jOOQ's `DefaultTransactionProvider`. That class tracks transaction nesting with a `Deque<Savepoint>` stored per `Configuration` - the first `begin()` call for a given `Configuration` is treated as the "top-level" transaction, and `commit()`/`rollback()` only actually touch the connection once that stack empties back to zero.

The relevant parts:

```java
// org.jooq.impl.DefaultTransactionProvider

private final boolean autoCommit(TransactionContext ctx) {
    Boolean autoCommit = (Boolean) ctx.configuration().data(DATA_DEFAULT_TRANSACTION_PROVIDER_AUTOCOMMIT);
    if (!TRUE.equals(autoCommit)) {
        autoCommit = connection(ctx).getAutoCommit();   // read ONCE, then cached
        ctx.configuration().data(DATA_DEFAULT_TRANSACTION_PROVIDER_AUTOCOMMIT, autoCommit);
    }
    return autoCommit;
}

private final void brace(TransactionContext ctx, boolean start) {
    ...
    boolean autoCommit = autoCommit(ctx);
    // Transactions cannot run with autoCommit = true. Change the value for the duration of a transaction
    if (autoCommit)
        connection.setAutoCommit(!start);
    ...
}

public final void commit(TransactionContext ctx) {
    ...
    if (savepoints.isEmpty()) {
        connection(ctx).commit();   // unconditional - runs regardless of the cached autoCommit value
        brace(ctx, false);
    }
}
```

When the connection arrives with `autoCommit` already `false`, `autoCommit(ctx)` caches `false`, and `brace()`'s `if (autoCommit) connection.setAutoCommit(...)` branch never fires - jOOQ correctly avoids touching a connection someone else already put into transactional mode. 

But `commit()` has no equivalent guard: once the savepoint stack for this `transactionResult()` call empties, it calls `connection(ctx).commit()` unconditionally. That's a real `Connection.commit()` at the JDBC level, which finalizes *everything* pending on that physical connection - not just the SQL the lock operation itself just ran, but anything the caller had already executed on that same connection object before handing it to jOOQ. jOOQ correctly infers "someone else is managing this connection's transaction boundaries" (that's why it leaves `autoCommit` alone) but then acts as if it's still the one responsible for ending that transaction anyway.

**Reproduction - silent premature commit of unrelated caller data**

I checked this against `postgres:18-alpine` with a few variations (`SilentPrematureCommitMatrixTest`, `DefaultJooqCommitBehaviorProbeTest`). The simplest case - a lock that's actually acquired - already shows it:

```java
// The caller's own, unrelated, not-yet-committed work - nothing to do with ShedLock.
try (Connection ambientConn = dataSource.getConnection()) {
    ambientConn.setAutoCommit(false);
    try (var stmt = ambientConn.createStatement()) {
        stmt.execute("INSERT INTO marker(id) VALUES (1)");
    }
    // not committed yet - a separate connection sees 0 rows here

    JooqStorageAccessor accessor = new JooqStorageAccessor(DSL.using(ambientConn, SQLDialect.POSTGRES));
    boolean acquired = accessor.insertRecord(
            new LockConfiguration(Instant.now(), "lock-1", Duration.ofMinutes(5), Duration.ZERO));
    // acquired == true
    // a separate connection now sees 1 row in marker, even though the caller never called commit()
}
```

```
marker count after SUCCESSFUL lock call: 1
```

What's more concerning is that a **losing** lock attempt behaves identically - `insertRecord()` returning `false` because another node already holds the lock doesn't stop the commit, since it happens regardless of whether the lock was actually acquired:

```java
// Someone else already holds "lock-2" (a completely independent, already-committed lock).
try (Connection winnerConn = dataSource.getConnection()) {
    new JooqStorageAccessor(DSL.using(winnerConn, SQLDialect.POSTGRES))
            .insertRecord(new LockConfiguration(Instant.now(), "lock-2", Duration.ofMinutes(5), Duration.ZERO));
}

try (Connection ambientConn = dataSource.getConnection()) {
    ambientConn.setAutoCommit(false);
    try (var stmt = ambientConn.createStatement()) {
        stmt.execute("INSERT INTO marker(id) VALUES (1)");
    }

    JooqStorageAccessor accessor = new JooqStorageAccessor(DSL.using(ambientConn, SQLDialect.POSTGRES));
    // This attempt LOSES the race - lock-2 is already held.
    boolean acquired = accessor.insertRecord(
            new LockConfiguration(Instant.now(), "lock-2", Duration.ofMinutes(5), Duration.ZERO));
    // acquired == false, but the marker row is committed anyway
}
```

```
marker count after FAILED (losing) lock call: 1
```

It also survives an explicit `rollback()`: if the caller calls `connection.rollback()` *after* the lock call, intending to undo everything, it's too late - the row was already committed by the lock call and the rollback has no effect on it:

```
marker count after caller's explicit rollback (post lock call): 1
```

This isn't specific to `insertRecord()` either - `extend()` and `unlock()` go through the same `transactionResult()` code path and commit just as unconditionally:

```
marker count after extend() : 1
marker count after unlock() : 2
```

And it doesn't matter how many unrelated rows are pending - three separate ones all get swept into the same commit that one would be:

```
marker count after lock call (3 unrelated rows pending before it): 3
```

The only case where the caller's row survives untouched is when the lock call runs on a connection obtained fresh from a `DataSource` rather than sharing the caller's own connection:

```
marker count after lock call on a SEPARATE connection: 0
```

- which is what the fix below relies on.

A batch job that opens one `Connection` manually and uses it, via jOOQ, to write several related rows meant to be committed or rolled back together at the end is a realistic way to hit this - it's a common pattern in code that manages its own transaction boundaries instead of relying on Spring's declarative `@Transactional` (which is also exactly how jOOQ's own getting-started documentation introduces `DSL.using(connection, dialect)`, before a `DataSource`-backed setup is introduced). If that same connection/`DSLContext` is also used partway through to acquire (or attempt to acquire) a ShedLock lock, everything written before that point is committed right then, regardless of what the batch decides afterward - nothing errors, nothing logs, and the batch's own later `rollback()` call quietly does less than its own code implies.

**Scope - which configurations are affected**

- **`DataSource`-backed `DSLContext`** (`DSL.using(dataSource, dialect)`, jOOQ's documented default): safe. `DefaultConnectionProvider`/`DataSourceConnectionProvider` pulls a fresh connection from the pool for every `transactionResult()` call, so there's never any of the caller's own pending work sharing that connection in the first place - nothing for the unconditional `commit()` to catch.

- **`DSLContext` via `TransactionAwareDataSourceProxy`** (jOOQ's documented recommendation for Spring integration, and what Spring Boot's `JooqAutoConfiguration` wires up by default) paired with
  - `NewTransactionRunner`/`SpringNewTransactionRunner`: also safe, for a specific reason. `TransactionAwareDataSourceProxy.getConnection()` doesn't hold on to one fixed `Connection` - every call checks Spring's `TransactionSynchronizationManager` (a thread-local registry) for whatever connection is currently bound to this thread's active transaction. 
  - When `SpringNewTransactionRunner` opens a `PROPAGATION_REQUIRES_NEW` transaction, Spring's `DataSourceTransactionManager` suspends the caller's existing thread-bound connection (just unbinds it from that registry, doesn't touch it) and binds a brand-new connection from the pool for the duration of the new transaction. Because the proxy re-checks the registry on every call, jOOQ's lock operation ends up running - and committing - entirely on that new, unrelated connection. 
  - When the callback returns, Spring commits that new connection's own transaction and rebinds the registry back to the caller's original connection, exactly as the caller left it. I verified this suspend-not-commit mechanism directly (for `JdbcTemplateStorageAccessor`, which uses the same Spring `REQUIRES_NEW` machinery): wrote a row inside an outer `PROPAGATION_REQUIRED` transaction, acquired a lock via `REQUIRES_NEW` in the middle of it, then rolled the outer transaction back - the row never committed (0 rows visible afterward), while the lock itself persisted independently (1 row). Since `SpringNewTransactionRunner` is the identical mechanism, the same guarantee applies to jOOQ.

- **`DSLContext` bound directly to a raw `Connection`** (`DSL.using(connection, dialect)` - jOOQ's own docs describe this as "a basic approach suitable for simpler scenarios," not the Spring-recommended one): this is where the bug reproduces. `DefaultConnectionProvider` here just holds the one `Connection` object it was constructed with and always returns that same instance - there's no registry to re-check, so `NewTransactionRunner` opening a Spring transaction elsewhere has nothing to redirect jOOQ to. jOOQ's own transaction handling has already run and committed on the caller's original connection by the time any such wrapping would matter.

**Suggested fix**

`JdbcStorageAccessor` and `JdbcTemplateStorageAccessor` both already require a `DataSource` for exactly this kind of isolation - pulling a genuinely independent connection instead of reusing whatever the caller's own `DSLContext`/`Connection` happens to be doing. I'd suggest the same for `JooqLockProvider`: an opt-in `JooqLockProvider(DSLContext, DataSource)` constructor, fully backward compatible with the existing single-arg one, where lock operations run on a fresh connection obtained directly from the `DataSource`. Since that connection is never shared with the caller's own connection, there's nothing of the caller's for the unconditional `commit()` to catch.

I looked at two alternatives first and ruled both out:
- Toggling `autoCommit` on the `DSLContext`'s *existing* connection (via `dslContext.configuration().connectionProvider()`) needs no new API, but per the JDBC spec, calling `setAutoCommit(true)` on a connection with an open transaction commits it - so it has the exact same forced-commit problem this issue describes, just triggered by my own code instead of jOOQ's.
- `NewTransactionRunner` only has something to redirect to when `dslContext`'s connection provider is itself registry-aware (`TransactionAwareDataSourceProxy`, as above). For a `DSLContext` bound to a raw `Connection`, there's no redirection target, so it can't help here.

I've opened #3684 with a working implementation and full test coverage (including everything described above) - would appreciate a look whenever you have time.

stlahxm:
### What

Adds an opt-in constructor overload, `JooqLockProvider(DSLContext, DataSource)`. When a `DataSource` is provided, every lock operation (`insertRecord`, `updateRecord`, `unlock`, `extend`) runs on a fresh connection obtained directly from it, instead of whatever connection the `DSLContext` happens to be using. The existing single-arg `JooqLockProvider(DSLContext)` constructor is unchanged and remains the default - this is purely additive.

### Why

See #3683 for the full writeup. Short version of the mechanism: `JooqStorageAccessor` runs lock operations through `dslContext.transactionResult(txCallable)`, which goes through jOOQ's `DefaultTransactionProvider`. That class caches whatever `autoCommit` value the connection had when it first saw it, and if that was already `false` (i.e. the connection already had an open transaction), it correctly leaves `setAutoCommit()` alone - but `commit()` has no matching guard:

```java
// org.jooq.impl.DefaultTransactionProvider
public final void commit(TransactionContext ctx) {
    ...
    if (savepoints.isEmpty()) {
        connection(ctx).commit();   // unconditional, regardless of the connection's original autoCommit state
        brace(ctx, false);
    }
}
```

So if the `DSLContext` given to `JooqLockProvider` is bound to a connection that's sharing an ambient transaction with something else the caller was doing, a single lock operation force-commits that connection - including whatever the caller had already written on it and not yet committed - whether or not the lock attempt itself succeeds.

I verified this holds across every operation and several edge cases (`SilentPrematureCommitMatrixTest`, `DefaultJooqCommitBehaviorProbeTest`):

```
successful lock call                                    -> caller's pending row: committed
failed/losing lock attempt                               -> caller's pending row: still committed
rollback() called on the caller's connection after the fact -> too late, no effect
3 unrelated pending rows before one lock call             -> all 3 committed
extend() / unlock()                                       -> same behavior as insertRecord()
lock call on a genuinely separate connection (DataSource-based) -> caller's row: untouched, rollback() still works
```

A realistic way to hit this: a batch job that opens one `Connection` manually and writes several related rows via jOOQ, meaning to commit-or-rollback them together at the end - a common pattern in code that manages its own transaction boundaries rather than relying on Spring's `@Transactional` (and also how jOOQ's own getting-started docs introduce `DSL.using(connection, dialect)`, before a `DataSource`-backed setup). 
If that same connection/`DSLContext` also gets used partway through to acquire a ShedLock lock, everything written before that point is committed right then regardless of what the batch decides afterward - no error, no log.

`JdbcStorageAccessor` and `JdbcTemplateStorageAccessor` don't have this problem because they already require a `DataSource` and never touch the caller's own connection at all. `NewTransactionRunner` (7.9.0, for #3577) fixes a related but narrower case: when `DSLContext`'s connection provider is itself Spring-registry-aware (`TransactionAwareDataSourceProxy`), Spring's `REQUIRES_NEW` can suspend the caller's connection and redirect jOOQ to a new one. 

But `DefaultConnectionProvider` (what backs `DSL.using(connection, dialect)`) just holds the one `Connection` it was constructed with and always returns that same instance - there's no registry for `NewTransactionRunner` to redirect through, so it can't help here. The only way to guarantee the caller's connection is never touched is to not use it in the first place, which needs a `DataSource`.

### Why not just toggle autocommit on the DSLContext's existing connection (no new API)?

I tried that first, since it needs no new constructor - `dslContext.configuration().connectionProvider()` gives back the `Connection`, which can be autocommit-toggled the same way `JdbcStorageAccessor` does. But per the JDBC spec, calling `setAutoCommit(true)` on a connection with an open transaction commits it - so this reproduces the identical forced-commit problem described above, just triggered by my own code instead of jOOQ's `DefaultTransactionProvider.commit()`. Pulling a genuinely separate connection is the only way to avoid it.

### Changes

- `JooqLockProvider`: new `JooqLockProvider(DSLContext, DataSource)` constructor, javadoc explaining when to use it.
- `JooqStorageAccessor`: takes an optional `DataSource`. When present, lock operations run via a connection pulled directly from `dataSource` (autocommit temporarily forced to `true` if needed, restored afterward, mirroring `JdbcStorageAccessor`), using a `DSLContext` derived from the original one's dialect/settings. When absent, behavior is unchanged from before this PR.
- New tests in `shedlock-provider-jooq`:
  - `SilentPrematureCommitMatrixTest` - the matrix above: successful/failed lock attempts, multiple pending rows, `extend()`/`unlock()`, and the separate-connection control case, all with the default (no-`DataSource`) constructor.
  - `JooqStorageAccessorDataSourceFixTest` - with the `DataSource` constructor: the caller's own pending work is never touched and still rolls back cleanly, and `lock_until` is computed from the real current time (the `DataSource`-based fix incidentally also resolves the timestamp-staleness issue from #3682/#3577, since it never shares a connection with an ambient transaction in the first place).
  - `AmbientTransactionStaleNowReproTest`, `AmbientTransactionDoubleAcquireReproTest` - carried over from the earlier #3682 investigation; document that the default constructor still reproduces the timestamp symptom too.

### Test plan

Full `shedlock-provider-jooq` module test suite, including all existing Postgres/MySQL/MariaDB/HSQLDB integration tests, passes unmodified alongside the new tests:

```
Tests run: 95, Failures: 0, Errors: 0, Skipped: 2
BUILD SUCCESS
```

(2 skipped tests are pre-existing/unrelated to this change.)

### Backward compatibility

Fully additive. The existing `JooqLockProvider(DSLContext)` constructor and its behavior are untouched; users who don't opt in see no change.

### DIFF
--- a/providers/jdbc/shedlock-provider-jooq/src/main/java/net/javacrumbs/shedlock/provider/jooq/JooqLockProvider.java
+++ b/providers/jdbc/shedlock-provider-jooq/src/main/java/net/javacrumbs/shedlock/provider/jooq/JooqLockProvider.java
@@ -13,6 +13,7 @@
  */
 package net.javacrumbs.shedlock.provider.jooq;
 
+import javax.sql.DataSource;
 import net.javacrumbs.shedlock.support.NewTransactionRunner;
 import net.javacrumbs.shedlock.support.StorageBasedLockProvider;
 import org.jooq.DSLContext;
@@ -24,5 +25,21 @@ public class JooqLockProvider extends StorageBasedLockProvider {
 
     public JooqLockProvider(DSLContext dslContext, NewTransactionRunner newTransactionRunner) {
         super(new JooqStorageAccessor(dslContext, newTransactionRunner));
+    }
+
+    /**
+     * Like {@link #JooqLockProvider(DSLContext)}, but every lock operation runs on a fresh
+     * connection obtained directly from {@code dataSource}, instead of whatever connection {@code
+     * dslContext} happens to be using. Unlike {@link #JooqLockProvider(DSLContext,
+     * NewTransactionRunner)}, this works even when {@code dslContext} is bound directly to a raw
+     * {@link java.sql.Connection} that may already have an ambient transaction open (e.g. {@code
+     * DSL.using(connection, dialect)}) - a construction {@code NewTransactionRunner} cannot help
+     * with, since it has no independent connection to switch to. Use this constructor if {@code
+     * dslContext} may be bound to a connection that is shared with, or participates in, a
+     * transaction the calling code manages itself, and {@code dslContext} isn't guaranteed to be
+     * Spring-transaction-aware.
+     */
+    public JooqLockProvider(DSLContext dslContext, DataSource dataSource) {
+        super(new JooqStorageAccessor(dslContext, dataSource));
     }
 }
--- a/providers/jdbc/shedlock-provider-jooq/src/main/java/net/javacrumbs/shedlock/provider/jooq/JooqStorageAccessor.java
+++ b/providers/jdbc/shedlock-provider-jooq/src/main/java/net/javacrumbs/shedlock/provider/jooq/JooqStorageAccessor.java
@@ -8,9 +8,11 @@ import static org.jooq.impl.DSL.localDateTimeAdd;
 import static org.jooq.impl.DSL.when;
 
 import java.io.Serializable;
+import java.sql.Connection;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.Map;
+import javax.sql.DataSource;
 import net.javacrumbs.shedlock.core.LockConfiguration;
 import net.javacrumbs.shedlock.support.AbstractStorageAccessor;
 import net.javacrumbs.shedlock.support.LockException;
@@ -21,11 +23,14 @@ import org.jooq.Field;
 import org.jooq.Record;
 import org.jooq.TableField;
 import org.jooq.TransactionalCallable;
+import org.jooq.impl.DSL;
 import org.jooq.types.DayToSecond;
+import org.jspecify.annotations.Nullable;
 
 class JooqStorageAccessor extends AbstractStorageAccessor {
     private final DSLContext dslContext;
     private final NewTransactionRunner newTransactionRunner;
+    private final @Nullable DataSource dataSource;
     private final Shedlock t = SHEDLOCK;
 
     JooqStorageAccessor(DSLContext dslContext) {
@@ -35,6 +40,18 @@ class JooqStorageAccessor extends AbstractStorageAccessor {
     JooqStorageAccessor(DSLContext dslContext, NewTransactionRunner newTransactionRunner) {
         this.dslContext = requireNonNull(dslContext, "dslContext can not be null");
         this.newTransactionRunner = requireNonNull(newTransactionRunner, "newTransactionRunner can not be null");
+        this.dataSource = null;
+    }
+
+    /**
+     * Runs every lock operation on a fresh connection obtained directly from {@code dataSource},
+     * bypassing {@code dslContext}'s own connection entirely - see {@link
+     * JooqLockProvider#JooqLockProvider(DSLContext, DataSource)}.
+     */
+    JooqStorageAccessor(DSLContext dslContext, DataSource dataSource) {
+        this.dslContext = requireNonNull(dslContext, "dslContext can not be null");
+        this.dataSource = requireNonNull(dataSource, "dataSource can not be null");
+        this.newTransactionRunner = TransactionCallback::execute;
     }
 
     @Override
@@ -83,6 +100,9 @@ class JooqStorageAccessor extends AbstractStorageAccessor {
     }
 
     private <T> T runInTransaction(TransactionalCallable<T> txCallable) {
+        if (dataSource != null) {
+            return runOnIndependentConnection(txCallable);
+        }
         try {
             @SuppressWarnings("unchecked")
             T result = (T) newTransactionRunner.runInNewTransaction(() -> dslContext.transactionResult(txCallable));
@@ -90,6 +110,35 @@ class JooqStorageAccessor extends AbstractStorageAccessor {
         } catch (Exception e) {
             throw new LockException(e);
         } catch (Throwable e) {
+            throw new LockException(e);
+        }
+    }
+
+    /**
+     * Pulls a genuinely independent connection from {@code dataSource} and runs {@code
+     * txCallable} there, never touching whatever connection {@code dslContext} itself is bound
+     * to. This is what makes it safe even when {@code dslContext} is bound directly to a raw
+     * {@link Connection} that may already have an ambient transaction (and other, unrelated
+     * caller work) pending on it.
+     */
+    private <T> T runOnIndependentConnection(TransactionalCallable<T> txCallable) {
+        try (Connection connection = dataSource.getConnection()) {
+            boolean originalAutoCommit = connection.getAutoCommit();
+            if (!originalAutoCommit) {
+                connection.setAutoCommit(true);
+            }
+            try {
+                DSLContext isolatedDsl = DSL.using(
+                        connection,
+                        dslContext.configuration().dialect(),
+                        dslContext.configuration().settings());
+                return isolatedDsl.transactionResult(txCallable);
+            } finally {
+                if (!originalAutoCommit) {
+                    connection.setAutoCommit(false);
+                }
+            }
+        } catch (Exception e) {
             throw new LockException(e);
         }
     }
--- a/providers/jdbc/shedlock-provider-jooq/src/main/java/net/javacrumbs/shedlock/provider/jooq/JooqStorageAccessor.java
+++ b/providers/jdbc/shedlock-provider-jooq/src/main/java/net/javacrumbs/shedlock/provider/jooq/JooqStorageAccessor.java
@@ -100,8 +100,9 @@ class JooqStorageAccessor extends AbstractStorageAccessor {
     }
 
     private <T> T runInTransaction(TransactionalCallable<T> txCallable) {
-        if (dataSource != null) {
-            return runOnIndependentConnection(txCallable);
+        DataSource ds = dataSource;
+        if (ds != null) {
+            return runOnIndependentConnection(ds, txCallable);
         }
         try {
             @SuppressWarnings("unchecked")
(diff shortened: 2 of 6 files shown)


lukas-krecan:
Replaced by https://github.com/lukas-krecan/ShedLock/pull/3725

PR #3725, opened 2026-09-03
#3683 Introduce JooqTransactionRunner

lukas-krecan:
To solve #3683


stlahxm:
@lukas-krecan Yes, this solves it. Verified against the raw-connection case, pending caller data now survives a rollback whether the lock succeeds or loses the race. Also confirmed dataSource keeps the original Settings intact. Different API shape than I suggested, but JooqTransactionRunner fits better next to NewTransactionRunner. Thanks for the quick turnaround 👍🏻

